PR #48, opened 2019-04-23
Replaced package mvtnorm by mvnfast

martinju:
We want to avoid requiring a very recent version of R to use our package. Currently, sampling is done with the mvtnorm package, requiring version 3.5, so we want to replace it. 

mvnfast is a computationally efficient package for sampling and evaluation of Gaussian distributions with few dependencies. 

**What I have done**:

- Replaced Gaussian sampling using the mvtnorm package with the mvnfast package for the copula and Gaussian methods.
- Updated DESCRIPTION and NAMESPACE with the new package.

**Notes:** 
- A basic test indicated about 15% reduction in computation time for the Gaussian method.
- The sampling may be parallelized with the mvnfast package, but I have currently disabled this option. Should handle that elsewhere.
- Considered using the MASS package, but that is considerably slower.


R/shapley.R
```diff
@@ -301,6 +301,7 @@ gaussian_transform <- function(x) {
 #' @return data.table with \code{noSamp_MC} (conditional) Gaussian samples
 #'
 #' @import condMVNorm
+#' @import mvnfast
 #' @export
```
nikolase90: This is not necessary. Remove the following lines (303 & 304):
```
#' @import condMVNorm
#' @import mvnfast
```
Note that you should regenerate documentation after doing this.

R/shapley.R
```diff
@@ -351,6 +352,7 @@ sample_copula <- function(given_ind, noSamp_MC, mu, Sigma, p, Xtest_Gauss_trans,
...
```
nikolase90: See comment above.

nikolase90:
@martinju Did you rerun the `REAME.Rmd`?

martinju:
> @martinju Did you rerun the `REAME.Rmd`?

Nope, will fix it.

R/shapley.R
martinju: Done. I thought this was the proper way to add packages that should be imported?

martinju:
> > @martinju Did you rerun the `REAME.Rmd`?
> 
> Nope, will fix it.

We don't use the Gaussian or copula methods there, so nothing is changed.

martinju:
@nikolase90: From my side this seems ready for merge now

R/shapley.R
nikolase90: It is not necessary when we're using `mvnfast::foo()`. And if it is only one function you'd like to import (and you're using it in multiple cases) it is actually better to use `@importFrom mvnfast foo`. Then you could change `mvnfast::foo()` to `foo()`. 

But I would recommend to use `mvnfast::foo()` due to two reasons: 
1) You'll immediately know that `foo()` is a function from `mvnfast`   
2) You avoid potential problems if there are cases where two (or more) packages have a function that shares the same name

Issue #22, opened 2019-04-04
Add linting of package


nikolase90:
The following files should be adjusted:
- [x] clustering.R
- [x] explanation.R
- [ ] features.R
- [ ] models.R
- [ ] observations.R
- [ ] plot.R
- [ ] predictions.R
- [ ] sampling.R
- [ ] shapley.R
- [x] shapr-package.R
- [ ] transformation.R
- [ ] utils.R

PR #52, opened 2019-05-02
Fixes all cases with non-snake_case variables

nikolase90:
Summarized
- Renames all cases where variables did not have `snake_case` names
- Updates documentation where there are new variable names
- Fixes a couple of cases where the line length was greater than 100
- Updates `.lintr`
- Adds a section in `.circleci/config.yml`where linting of package is done. Note that circleCI will fail if the package contains lints 
- Updates  `.circleci/config.yml` so that we install the github version of `lintr` and not the version that is available on CRAN

Fixes https://github.com/NorskRegnesentral/shapr/issues/22 & https://github.com/NorskRegnesentral/shapr/issues/10.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,9 @@ jobs:
           name: Install devtools
           command: R -e 'install.packages("devtools")'
       - run:
+          name: Install lintr
+          command: R -e 'devtools::install_github("jimhester/lintr")'
+      - run:
           name: Install dependencies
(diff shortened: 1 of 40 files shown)


nikolase90:
Agreed! You're very welcome to find additional cases, and I could fix them.

nikolase90:
Considering we paused the project for a while after opening this PR, I'm closing this one. A bit to risky to do this before we have any kind of tests in place.

PR #74, opened 2019-07-10
Adds documentation and test for functions in R/extra.R

JensWahl:


tests/testthat/test-sample_combination.R
```diff
@@ -0,0 +1,56 @@
+library(shapr)
+
+context("test-sample_combinations.R")
+
+test_that("Test function on extra.R", {
```
nikolase90:
```suggestion
test_that("Test sample_combinations", {
```

R/extra.R
```diff
@@ -2,13 +2,20 @@
 #' getting the same observation twice. Need to improve this help file.
 #'
 #' @inheritParams global_arguments
+#'
+#' @param nTrain Number of training observations to sample from.
+#'
+#' @param nTest Number of test observations to sample from.
+#'
+#' @param nosamp Number of samples
+#'
```
nikolase90: Could you add which type each of the param's is? 

I.e. if nTrain is should be a positive integer vector, do 
```
#' @param nTrain Positive integer. Number of training observations to sample from.
```

R/extra.R
```diff
@@ -2,13 +2,20 @@
...
 #' @param separate Logical indicating whether the train and test data should be sampled separately
 #' or in a joint sampling space. If they are sampled separately (which typically would be used when
 #' optimizing more than one distribution at once) we sample with replacement if more samples than
 #' training data. Not optimal, but for now fine if careful when using more samples than the number
 #' training observations while at the same time doing optimization over every test observation.
 #'
-#' @return Numeric
+#' @return A data.frame containing \code{nosamp} rows of resampled train and test observations.
```
nikolase90:
```suggestion
#' @return Data.frame. Contains \code{nosamp} rows of re-sampled train and test observations.
```

tests/testthat/test-transformation.R
```diff
@@ -0,0 +1,24 @@
+library(shapr)
```
nikolase90: You should have this file in a separate PR. Don't know if this was intentional or not.

PR #75, opened 2019-07-10
Removes scale_data function in R/utils.R

nikolase90:
We don't use `scale_data()` anymore. If we at some point in the future need the function we'll just revert this change.


nikolase90:
@JensWahl Hopefully you haven't written any tests for this function?

JensWahl:
No, there are no tests for this function.

PR #76, opened 2019-07-10
Updates sample_combinations in R/extra.R

nikolase90:
@JensWahl Could you confirm that this is OK? 

This is basically just to reduce the number of lines, and make the code a bit more readable.


R/extra.R
```diff
@@ -21,52 +21,39 @@
...
+      x = nTrain,
+      size = nosamp,
+      replace = ifelse(nosamp < nTrain, FALSE, TRUE)
+    )
+
+    # Sample test data
+    samp_test <- sample(
+      x = nTest,
+      size = nosamp,
+      replace = ifelse(nosamp < nTrain, nosamp > nTest, TRUE)
```
JensWahl: Why are we checking `nosamp < nTrain`?

nikolase90: We did that previously as well?

nikolase90:
So I think you need to have write-access to give a valid review, I'll fix that tomorrow.

PR #80, opened 2019-07-11
Changes variable names in sample_combinations()

nikolase90:
Motivation behind the changes of variable names; 
- We don't want a combination of `camelCase` and `snake_case` in our code. We've decided to use `snake_case` and hence I'm changing `nTrain` and `nTest` to `ntrain` and `ntest`. 
- When reading `nosamp` I thought it indicated "No sampling", hence I changed it to `nsamples`. 
- I think `joint_sampling` is more informative than `separate`. To understand the latter you have to read the documentation. 

@JensWahl What do you think? Does this makes sense?

### DIFF
--- a/R/extra.R
+++ b/R/extra.R
@@ -9,20 +9,21 @@
 #'
 #' @param nsamples Positive integer. Number of samples.
 #'
-#' @param separate Logical indicating whether the train and test data should be sampled separately
-#' or in a joint sampling space. If they are sampled separately (which typically would be used when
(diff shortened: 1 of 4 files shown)


JensWahl:
I agree, but should we change the name from `nTrain` to `n_train` etc?

nikolase90:
I thought about it, but I feel like we should stick with `ntrain` and `ntest`. Perhaps only use `_` when there are 2 or more letters in first part?

JensWahl:
Sounds good!

PR #84, opened 2019-08-20
Adds link to paper in readme.

martinju:
Adds link to paper from readme. Easier to reach then.


nikolase90:
Any idea why there are so many changes when you only added a link?

martinju:
No, unfortunately. I just changed the rmd file.

nikolase90:
Looks like it was just of version thing - I re-knitted the file and it fixed most of it. I also worth nothing that we should use a new line more often in `README.Rmd`, since it looks like we changed all of line 32 in this case (we only changed a minor part).

PR #85, opened 2019-08-20
Adds vignette

### LINKED ISSUE #17: Add vignette of methodology/package
I made a template with https://github.com/NorskRegnesentral/shapr/tree/nikolai/add_vignette. You could continue pushing to this branch once you have the time work on this. 

Some examples of vignettes: 
- https://github.com/thomasp85/lime
- https://github.com/osorensen/BayesMallows

martinju:
Fixes #17. 

Mainly Camilla's work. I went through it and fixed the text and figures a bit.

Still missing:

- [x] Comparison with original SHAP implementation.
- [x] Mention various features of the implementation. I.e. briefly what settings are available to the user in the package.


martinju:
@nikolase90 Is it forbidden to include the compiled html file under /vignettes? I see that you've added .html to .gitignore. I find it convenient to be reach the vignette directly from github without needing to compile the .Rmd file myself.

martinju:
What remains in my opinion is to mention and briefly explain what settings the user can adjust to do different things. This should be done referring to the updated version of the package #83. @Camiling, can you do this on Tuesday?

nikolase90:
@martinju We could probably add the .html file of the vignette. It's just a default setting to ignore it. I'm not sure if you will be able to look at the file without downloading it though, since I don't think Github renders html code. Anyways, feel free to add the html as well :+1:

martinju:
Turns out no html file is produced by knitr for the html_vignette(), when clicking "knit" in Rstudio, which is how I do things, see 
https://stackoverflow.com/questions/23683994/knit-html-does-not-save-html-in-vignettes

We could always do render(FILE.Rmd), but I guess that will be forgotten, and the .html file will easily be outdated. If we build a simple pkgdown website (#68 ) some time autumn we could rather include it there (does not have to be fancy at all). I therefore suggest to just to keep only the .Rmd file for now.

nikolase90:
Yeah, I think perhaps the website is the best solution. I will take 10 minutes to get that up and running.

martinju:
Sounds good. Lets do that within the next month then.

PR #96, opened 2019-08-29
Moves some packages from Imports to Suggets

nikolase90:


nikolase90:
Yep. Should only be present in `R/clustering.R`. I couldn't figure out if the package `cluster` was used anywhere though. Do you if that is a package we use? It's listed under `Imports`

martinju:
Check `R/clustering` and does not seem to be in use. I think we can delete it.

nikolase90:
Ok. Removed it with 7347542.

PR #97, opened 2019-08-29
Adds mgcv to Imports

nikolase90:
Forgot to add this under #89

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,6 @@ Imports:
     stats,
     data.table,
     Rcpp (>= 0.12.15),
-    mgcv,
     cluster,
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -26,6 +26,7 @@ Imports:
     stats,
     data.table,
     Rcpp (>= 0.12.15),
+    mgcv,
```
martinju: Shouldn't this be under suggests along with ranger, xgboost etc?

nikolase90: Indeed! Will fix.

nikolase90: Fixed with 525f46e

PR #99, opened 2019-09-02
Adds new functions, and some refactoring of code

nikolase90:


R/explanation.R
```diff
@@ -0,0 +1,330 @@
...
+  # Add arguments to explainer object
+  explainer$x_test <- as.matrix(x)
+  explainer$approach <- approach
+  explainer$type <- type
+  explainer$fixed_sigma_vec <- fixed_sigma_vec
+  explainer$AICc_no_samp_per_optim <- AICc_no_samp_per_optim
+  explainer$AIC_optim_max_eval <- AIC_optim_max_eval
+  explainer$AIC_optim_startval <- AIC_optim_startval
+  explainer$w_threshold <- w_threshold
+  explainer$n_samples <- AICc_no_samp_per_optim
```
martinju: I think this might be wrong. The number of samples used in the AICc optimization need not be the same as the number of samples that we use for the empirical method in the end... They may very well be two differnet numbers. For now I don't manage to see how explainer$n_samples gets passed for the empirical method though. Please take look at this. Otherwise it looks good to me for now.

nikolase90: You're right. I'll fix it.

nikolase90: Fixed with 11773f4.

martinju: Perfect!

PR #100, opened 2019-09-03
Fixes cases where we should not export the functions.

nikolase90:
Currently we export all functions, but this is not something we want to do. In this PR I modify this for all the functions in `R/features.R`, `R/sampling.R` and `R/transformation.R`. This is done by removing `@export` and replacing it with `@keywords internal`. 

We want to export like 5-10 functions, which looks much cleaner when a user uses `help(package = "shapr")`.


nikolase90:
@martinju Do you have time to approve this before you leave? Then I'll get a PR opened for the rest later tonight.

nikolase90:
Yep, totally agree. I just started with these (since the functions are tested and doesn't mess to much with the upcoming pull request).

PR #101, opened 2019-09-03
Adds model_type, and fixes some documentation in predict_model

nikolase90:


nikolase90:
See b668d03. We can do another review on the documentation once the code is in place.

Issue #54, opened 2019-05-08
Simplify user modification of the empirical approach


martinju:
Currently the user must supply the kernelshap function with a list specifying all details of the empirical condtional approach. There are default values for the full list, but if the user just needs to change one of these values, he needs to specify all the other default values manually, which is very inconvenient. Thus, if an element of the list is not supplied, default values should be used.

I think the easiest way out of this is to start with the default list in the top of the function and modify that with the elements in the list supplied in the function call.

PR #102, opened 2019-09-10
Fixes combined version

nikolase90:
Fixes https://github.com/NorskRegnesentral/shapr/issues/54.


R/explanation.R
```diff
@@ -8,9 +8,9 @@
 #' @param explainer An \code{explainer} object to use for exaplaining the observations.
 #' See \code{\link{shapr}}.
 #'
-#' @param approach String or list. If a String, the approach to use when computing the Shapley
-#' values. Equal to 'empirical', 'gaussian' or 'copula'. If a list, the elements in the list
-#' refers to the rows in \code{x} that ought to be included in each of the approaches.
+#' @param approach Character. Note that \code{1 <= length(approach) <= n_features}, where
+#' \code{n_features} where equals the total number of features in the model. All elements should
```
martinju: typo: where \code{n_features} *where* equals

R/explanation.R
```diff
@@ -8,9 +8,9 @@
 #' @param explainer An \code{explainer} object to use for exaplaining the observations.
 #' See \code{\link{shapr}}.
 #'
-#' @param approach String or list. If a String, the approach to use when computing the Shapley
-#' values. Equal to 'empirical', 'gaussian' or 'copula'. If a list, the elements in the list
-#' refers to the rows in \code{x} that ought to be included in each of the approaches.
+#' @param approach Character. Note that \code{1 <= length(approach) <= n_features}, where
```
martinju: Suggests:  @param approach Character (vector). Note that ....

R/explanation.R
```diff
@@ -8,9 +8,9 @@
...
+#' \code{n_features} where equals the total number of features in the model. All elements should
+#' either be \code{gaussian}, \code{copula} or \code{empirical}. See details for more information.
```
martinju: shouldn't these be in quotes? "gaussian", "copula" etc.

R/explanation.R
```diff
@@ -20,10 +20,10 @@
 #'
 #' @param seed Positive integer. If \code{NULL} a random seed will be used.
 #'
-#' @param ... Additional arguments passed to \code{explain.empirical}, \code{explain.gaussian} or
-#' \code{explain.copula}.
+#' @param ... Additional arguments passed to \code{\link{prepare_data}}
 #'
 #' @details
+#' TODO: Add information about approach.
```
martinju: Should be added now. Currently it is impossble to guess what happens when length(approach) != n_features

R/observations.R
```diff
@@ -52,11 +55,23 @@ prepare_data <- function(x, ...) {
...
+    x$x_train,
+    x$x_test,
+    x$X$features[index_features]
+  )
 
   # Setup
+  if (!is.null(seed)) set.seed(seed)
   n_col <- nrow(x$x_test)
-  no_empirical <- nrow(x$S)
+  no_empirical <- nrow(x$S[index_features, ])
```
martinju: I think you need drop=F here to handle cases where length(index_features)=1. Shouldn't a test have captured such a problem?

R/observations.R
```diff
@@ -87,13 +102,13 @@ prepare_data.empirical <- function(x, seed = 1, n_samples = 1e3, ...) {
...
     if (kernel_metric == "independence") {
-      D <- D[sample(nrow(D)), ]
+      D <- D[sample.int(nrow(D)), ]
       h_optim_vec <- mean(D) * 1000
     }
 
     val <- t(t(-0.5 * D) / h_optim_vec^2)
     W_kernel <- exp(val)
-    S <- x$S
+    S <- x$S[index_features, ]
```
martinju: I think the drop=F applies here as well.

R/explanation.R
nikolase90: Fixed with 50295a1.

R/explanation.R
nikolase90: Fixed with dee3db7.

R/explanation.R
nikolase90: Fixed with 2fa0499.

R/observations.R
nikolase90: `index_features` should never be an element less than 2. I'll need to make an small change to make sure that this is the case.

nikolase90: This should fix it. c7fe376

PR #103, opened 2019-09-20
Updates README and adds plot.shapr instead of plot_kshap 

nikolase90:


R/shapley.R
```diff
@@ -45,29 +45,24 @@ weight_matrix <- function(X, use_shapley_weights_in_W = TRUE, normalize_W_weight
...
 #' (i.e. the explanatory variables). Note that the response variable should not be part of
 #' \code{x}.
 #'
-#' @param n_combinations Integer. The number of feature combinations to sample. If ´\code{NULL},
-#' the exact method is used and all combinations are considered. The maximum number of
-#' combinations equlas \code{2^p}, where \code{p = ncol(x)} (total number of explanatory variables).
+#' @param model The model whose predictions we want to explain.
 #'
-#' @details
-#' TODO: Write about missing values and categorical variables
```
martinju: Should this be done? Or is it mentioned elsewhere that we only deal with continuous variables ?

nikolase90: It is mentioned in `README.md`, but we should add some additional checks on the data. I.e. if the user pass training data that is not numerical or contains missing values, a meaningful error message should be returned.

PR #106, opened 2019-09-24
Fixes some warnings/notes when running R CMD CHECK

nikolase90:


R/predictions.R
```diff
@@ -11,7 +11,7 @@
 #' TODO: Write details about how this is done (reference to paper)
 #'
 #' @examples
-#' TODO: Add simple examples
+#' # TODO: Add simple examples
```
martinju: Extra #?

nikolase90: The code below `@examples` should be actual code, and `TODO: Add simple examples` will error, but `# TODO: Add simple examples` will work fine (since it's just commented code).

PR #107, opened 2019-09-24
Renames some of the arguments/variables in R/sample.R

nikolase90:
This is done due to two different things; 
1) Linting 
2) Should line up with arguments in `explain` functions


R/sampling.R
```diff
@@ -1,98 +1,101 @@
...
+#' @return data.table with \code{n_samples} (conditional) Gaussian samples
 #'
 #' @keywords internal
 #'
 #' @examples
 #' # TODO: Add simple example
 #'
 #' @author Martin Jullum
-sample_copula <- function(given_ind, noSamp_MC, mu, Sigma, p, Xtest_Gauss_trans, Xtrain, Xtest) {
+sample_copula <- function(index_mu, n_samples, mu, sigma, p, x_test_gaussian, x_train, x_test) {
```
martinju: What is the intuition behind the name change given_ind -> index_mu ? What is mu? I think given_ind is easy to understand along with dependent_ind.

nikolase90: My intuition behind `index_mu` is that it represents indices in `mu` (the other argument). Basically   easier to understand the input without looking at the actual code inside the function.

nikolase90:
> I am not so happy with changing Sigma to sigma. sigma_mat is OK, not to confuse it with a) the fixed_sigma in the empirical approach and b) a single standard deviation. This goes for all the functions here.
> 
> Also, see comment on index_mu.

I agree :+1:

R/sampling.R
nikolase90: @martinju Changed to `index_given`

nikolase90:
@martinju Everything should be fixed now (I think). It turns out we actually used `cov_mat` instead of `Sigma` in the higher level functions, so I went for that name.

PR #111, opened 2019-10-10
Renames argument in feature_combinations and adds checks for train/test-data

nikolase90:
Fixes the following three issues; 
- Fixes documentation for `type` in `explain.combined`
- Renames argument noSamp as `n_combinations` in `feature_combinations`
- Adds check for single prediction in `shapr` and `explain` (want to check that correct features are passed into the functions)
- Adds check that uses `exact = FALSE` if `m` is greater than a given number 
- Adds `MASS` under `Suggests` in `DESCRIPTION`


R/features.R
```diff
@@ -16,22 +16,33 @@
 #' @keywords internal
 #'
 #' @author Nikolai Sellereite, Martin Jullum
-feature_combinations <- function(m, exact = TRUE, noSamp = 200, shapley_weight_inf_replacement = 10^6, reduce_dim = TRUE) {
+feature_combinations <- function(m, exact = TRUE, n_combinations = 200, shapley_weight_inf_replacement = 10^6, reduce_dim = TRUE) {
+
+  # Force user to use a natural number for n_combinations if m > 10
```
martinju: Suggest using m > 12 instead. This is 8000 and should be perfectly fine. 

Suggesting to adjust the text to 

Due to the computational complexity, we recommend setting n_combinations to 10 000 instead of the exact method when the number of features is larger than 12. To force use the exact method, set n_combinations equal to 2^m, where m is the number of features.

martinju: Suggest using m > 12 instead. This is 8000 and should be perfectly fine.

Suggesting to adjust the text to

Due to the computational complexity, we recommend setting n_combinations to 10 000 instead of the exact method when the number of features is larger than 12. To force use the exact method, set n_combinations equal to 2^m, where m is the number of features.

R/features.R
```diff
@@ -16,22 +16,33 @@
...
-    stop("Currently we are not supporting cases where m > 30.")
+    stop("Currently we are not supporting cases where the number of features is greater than 30.")
   }
 
-  if (!exact && noSamp > (2^m - 2) && !reduce_dim) {
-    noSamp <- 2^m - 2
-    cat(sprintf("noSamp is larger than 2^m = %d. Using exact instead.", 2^m))
+  if (!exact && n_combinations > (2^m - 2) && !reduce_dim) {
+    n_combinations <- 2^m - 2
+    cat(sprintf("n_combinations is larger than 2^m = %d. Using exact instead.", 2^m))
```
martinju: It might be confusing when you are told to set n_combinations = 2^m, then this warning tells you that you have used MORE than 2^m.
Suggests:
cat(sprintf("n_combinations is larger than or equal to 2^m = %d. Using exact instead.", 2^m))

R/features.R
nikolase90: Fixed with fa8ec7b.

R/features.R
nikolase90: Fixed with fa8ec7b.

PR #119, opened 2019-10-28
Bugfix in shapr

martinju:
Fixes bug when x in shapr is a data.table. 
Typically worked OK before as x_train accidentally was a global variable.
Also, it is probably smart to not do x_train <- x, but rather make x_train a copy of x (as as.data.table(x) does when x is a data.table), since changing x afterwords should not affect explainer$x_train.

Example script
```
library(shapr)

data("Boston", package = "MASS")

y_var <- "medv"

x_train0 <- as.matrix(tail(Boston[, x_var], -6))
y_train <- tail(Boston[, y_var], -6)


model <- xgboost::xgboost(
  data = x_train0,
  label = y_train,
  nround = 20,
  verbose = FALSE
)
aa = data.table::as.data.table(x_train0)
# THIS USED TO THREW AN ERROR
explainer <- shapr(aa, model) 

all.equal(explainer$x_train,aa) # Should be identical
#[1] TRUE

# Changing aa should not change explainer$x_train
aa[,lstat:=1]
#explainer$x_train
all.equal(explainer$x_train,aa)
#[1] "Column 'lstat': Mean relative difference: 0.9215172"

```

### DIFF
--- a/R/shapley.R
+++ b/R/shapley.R
@@ -83,10 +83,8 @@ shapr <- function(x,
   # Checks model and features
   explainer$p <- predict_model(model, head(x))
 
-  # Create data.table --------------
-  if (!data.table::is.data.table(x)) {
-    x_train <- data.table::as.data.table(x)
-  }
+  # Converts to data.table, otherwise copy to x_train  --------------
+  x_train <- data.table::as.data.table(x)
 
   # Get all combinations ----------------
   dt_combinations <- feature_combinations(


R/shapley.R
```diff
@@ -83,10 +83,8 @@ shapr <- function(x,
   # Checks model and features
   explainer$p <- predict_model(model, head(x))
 
-  # Create data.table --------------
-  if (!data.table::is.data.table(x)) {
-    x_train <- data.table::as.data.table(x)
-  }
+  # Converts to data.table, otherwise copy to x_train  --------------
+  x_train <- data.table::as.data.table(x)
```
nikolase90: Nice catch!

Could we instead add an else statement like this? 
```
# Create data.table --------------
if (!data.table::is.data.table(x)) {
    x_train <- data.table::as.data.table(x)
} else {
    x_train <- data.table::copy(x)
}
```

martinju: I actually did that first, then realized that as.data.table.data.table(x) does almost exactly this, so wouldn't my suggestion just simplify the code?

https://github.com/Rdatatable/data.table/blob/50f815780297aaea5e30e9fe3c787d931957fe21/R/as.data.table.R#L239-L248

nikolase90: Awesome. Didn't know that. Go for it!

PR #122, opened 2019-10-30
Fix randomness and incorrect class in tests for sampling

martinju:
I discovered (the hard way) that at least one test may fail due to randomness when sampling running the tests (that ensure_condcov_symmetry should give different answers when true and false). Thus, set a fixed seed before any sampling in all tests of the sampling functions to avoid this.

Also fixes incorrect input format for x_test and x_test_gaussian in the sample_gaussian. Should be matrices (or data frames), but was vector in the tests. The same applies to sample_copula, but fixing it there introduced other issues.


martinju:
@nikolase90 
For e.g. set.seed(13) it throws an error for the sample_gaussian test with condcov:

```
set.seed(13) # Setting seed to avoid sampling a cov_mat which is exactly symmetric in certain dimensions,
# such that ensure_condcov_symmetry would not affect the results
cov_mat <- cov(matrix(rnorm(n_samples * m), n_samples, m))
x_test <- MASS::mvrnorm(2, mu, cov_mat)[1,,drop=F] # sample_gaussian requires x_test to be a matrix

# Example 3 -------------
# Check that ensuring conditional covariance matrix symmetry is FALSE by default.
index_given <- 4:7
set.seed(1)
res3.1 <- shapr:::sample_gaussian(index_given, n_samples, mu, cov_mat, m, x_test, ensure_condcov_symmetry = F)
set.seed(1)
res3.3 <- shapr:::sample_gaussian(index_given, n_samples, mu, cov_mat, m, x_test, ensure_condcov_symmetry = T)


testthat::expect_false(sum(res3.1 != res3.3) == 0) # Expect different results
# Error: sum(res3.1 != res3.3) == 0 isn't false.
```

Thus, I think it is important to set a seed for such a test. Of 100 different seeds it failed 11 times.

nikolase90:
Nice. There are a few other problems with the tests, so I'll add some additional changes.

nikolase90:
@martinju I removed the part with `ensure_condcov_symmetry` in the code base. Basically, in `condMVNorm::condMVN` there is an argument called `check.sigma` which says the following; 
> logical; if TRUE, the variance-covariance matrix is checked for appropriateness (symmetry, positive-definiteness). This could be set to FALSE if the user knows it is appropriate.

The default value equals `TRUE`, and it will fail if the matrix is not symmetric/positive-definite. I've also updated the tests appropriately.

martinju:
Regarding `ensure_condcov_symmetry`: This is exactly the problem. If it isn't exactly symmetric then condMVN will fail and thus explain will fail. 

The problem is that even if the full covariance matrix is symmetric, when computing conditional covariance matrices for some of the features, the conditional covariance matrix will in some rare cases not become exactly symmetric (due to numerics), and then condMVN will fail and we don't get a result. On the other hand, always doing what `ensure_condcov_symmetry=TRUE` does slows down the computations in high dimensions significantly, so you don't want to use it unless you really need to. Thus, unless you can dig up a better solution, I think we should keep that part as before.

nikolase90:
1. The result of `condMVN` is not dependent on whether `ensure_condcov_symmetry` is included or not. 
2. The problem I have with `ensure_condcov_symmetry` is that it modifies the conditional co-variance matrix regardless of whether the conditional co-variance matrix is symmetric and positive-definite. 

I'm fine with keeping the argument, but the tests and code should test what the argument actually should do. If for a given example setting `ensure_condcov_symmetry` equal to `TRUE` or `FALSE` should return the same result if the co-variance matrix is symmetric and positive-definite. That is not the case now.

martinju:
1. You are right. I mistook `condMVNorm::condMVN `for `mvnfast::rmvn`. The latter fails if the `sigma` argument is not exactly symmetric. 
2. Yes, we could always make a test of symmetry first. 
3. Agree, the current test of this is not worth much.

nikolase90:
Goodie! Added the argument again, and modified the code so that we only use `Matrix::symmpart` if the conditional co-variance matrix is not symmetric and positive-definite.

martinju:
Ok, good. Since you made some changes I will look it over once more before merging.

nikolase90:
Sounds good. Could you remove the files when you merge it?

PR #126, opened 2019-10-31
Renames x_test in prepare_data.copula

nikolase90:
Renamed `x_test` to `x_test_gaussian` in `prepare_data.copula`. 

In addition to this there's also a bug fix for the case when you only pass a single test observation. See the lines where we did `if (is.null(dim(x))`. 

To streamline this, the variables was also changed in `explain.copula`.


nikolase90:
@martinju Agree.

PR #127, opened 2019-10-31
Updates url and R-version in DESCRIPTION

nikolase90:
`mvnfast` (or in fact `mvtnorm`) is not supported for R < 3.5.0. I think we should just go for R>=3.5.0. It's less work to maintain, and it is almost 2 years old.

Also fixes an issue with a link in `README.md`. There was a misspecified link in the contribution paragraph.


martinju:
Agree. I saw there was a misspecified link when reviewing, but just assumed it was because the branch was master.

PR #128, opened 2019-10-31
Tests for shapley.R and explanation.R

martinju:
Both tests rely on an xgboost model (not fitted in the test to ensure consistency across different xgboost versions) and a small subset of the Boston data.

The tests check if the complete output from shapr and explain is the same as before. 

For the explain function the 18 different examples from `test-refactor.R`. All examples are tested in the same call, but the error messages specifies where the error is for easy debugging.


nikolase90:
Smooth!

tests/testthat/test-explanation.R
```diff
@@ -0,0 +1,105 @@
...
+  empirical_settings <- list(type = "independence", fixed_sigma_vec = 0.1, w_threshold = 0.95)
+  Ex.list[[3]] <- explain(x_test, explainer, approach = "empirical", prediction_zero = mean(y_train), type = "independence")
+
+  # Ex 4: Explain predictions (empirical, fixed sigma)
+  empirical_settings <- list(type = "fixed_sigma", fixed_sigma_vec = 0.1, w_threshold = 0.95)
+  Ex.list[[4]] <- explain(x_test, explainer, approach = "empirical", prediction_zero = mean(y_train), type = "fixed_sigma")
+
+
+  # Ex 5: Explain predictions (empirical, AICc)
+  empirical_settings <- list(
```
nikolase90: Could you just remove everything that is `empirical_settings`. Only used in the old code.

martinju: Haha, good point! Fixed in feb4f30.

nikolase90:
Could you add a script, i.e. `inst/script/test_data.R` that generates the all the files? Could be nice to have if we at some point regenerate the files.

martinju:
Sure. The `expect_known_value` does this if the file don't exist, such that all that is needed is perhaps a script for generating the xgboost model?

nikolase90:
Ah, ok. So if we wanted to update the results at some point we would just delete the file that is currently there, right? And then the tests would write a new file?

martinju:
Exactly. If there is no file the test will pass, but give a warning.  I will make a script creating the model, though.

nikolase90:
Question; 
I pulled the branch, installed the package, and ran devtools::test(). Everything is OK. 

Then I deleted `explanation_explain_obj_list.rds`. Did the same steps as above, and it fails. Any idea why?

nikolase90:
@martinju Do you want me to take a look?

martinju:
Ok, this has to be a versioning issue. I upgraded to CRAN version of xgboost (0.90.2), and the tests fail to predict with xgboost (i.e. completely different issue). 
@nikolase90 Maybe you can try to pull, delete the xgboost and test objects, then create the xgboost object, and then run devtools::test() twice and push the new files? I have a feeling that might just work.

nikolase90:
Sure!

martinju:
And, yes, feel free to try to debug on your end if that doesn't do the trick.

martinju:
Woohoo, test passed!

PR #133, opened 2019-11-05
Removes deprecated code 

nikolase90:
Note that with 58b6419 I changed the model (since we use a linear model in our tests), and confirmed that the new functionality returns the same results as the old.


nikolase90:
Woop woop!

Issue #138, opened 2019-11-21
Fix redundant arguments


martinju:
A list of potentially redundant function arguments that we could remove to simplify code.

- [x] We may delete the `reduce_dim` argument of `feature_combinations` (always set it to TRUE)
- [ ] If we delete the `reduce_dim` argument, we probably don't need the `use_shapley_weights_in_W` argument in `weight_matrix`, as the no-column is always 1 (check!!!), and then we can simplify this code by always using the `shapley_weight` column only as the weights.
- [x] Remove `type` in `inv_gaussian_transform`

nikolase90:
Fixed with #141 and #142 .

PR #141, opened 2019-11-21
Removes redundant arguments

nikolase90:
Fixes https://github.com/NorskRegnesentral/shapr/issues/138


nikolase90:
Question: Should I drop the `no` column (everywhere it is), and update  `tests/testthat/test_objects/shapley_explainer_obj.rd ` in this PR as well?

martinju:
> Question: Should I drop the `no` column (everywhere it is), and update `tests/testthat/test_objects/shapley_explainer_obj.rd ` in this PR as well?

From what I could see, we have no use for it (so yes, remove it), but please check carefully that you agree with me that no will always be 1, and that we do not need it for something else. I didn't do a thorough search.

R/features.R
```diff
@@ -135,11 +133,12 @@ feature_not_exact <- function(m, n_combinations = 200, weight_zero_m = 10^6, red
   X[, ind := NULL]
 
   # Set column order and key table
-  nms <- c("ID", "features", "nfeatures", "N", "shapley_weight", "no")
+  nms <- c("ID", "features", "nfeatures", "N", "shapley_weight")
   data.table::setcolorder(X, nms)
   data.table::setkey(X, nfeatures)
   X[, ID := .I]
   X[, N := as.integer(N)]
+  X[, no := 1] # TODO: Remove this after updating test objects
```
martinju: I don't understand why you have put this here. 
I think the crucial thing to check is that with the previous code and reduce_dim = TRUE, no will be equal to rep(1,nrow(X)) both when there are indeed duplicates and when there are no duplicates. I don't know if you have checked that, but if not, I think it is worth checking that with an example.

nikolase90: Oh, this was just to avoid that the tests failed.

R/features.R
```diff
@@ -54,15 +52,17 @@ feature_combinations <- function(m, exact = TRUE, n_combinations = 200,
...
+    exact <- TRUE
     cat(sprintf("n_combinations is larger than or equal to 2^m = %d. Using exact instead.", 2^m))
   }
 
   if (exact) {
     dt <- feature_exact(m, weight_zero_m)
   } else {
-    dt <- feature_not_exact(m, n_combinations, weight_zero_m, reduce_dim)
+    dt <- feature_not_exact(m, n_combinations, weight_zero_m)
+    dt[, p := NULL]
```
martinju: Is there a reason you did not remove this within feature_not_exact?

R/features.R
```diff
@@ -112,34 +111,33 @@ feature_not_exact <- function(m, n_combinations = 200, weight_zero_m = 10^6, red
...
-  nms <- c("ID", "nfeatures", "features", "no")
-  data.table::setcolorder(X, nms)
 
   # Add shapley weight and number of combinations
-  X[, shapley_weight := weight_zero_m]
+  X[c(1, .N), shapley_weight := weight_zero_m]
   X[, N := 1]
   X[between(nfeatures, 1, m - 1), ind := TRUE]
-  X[ind == TRUE, shapley_weight := p[nfeatures]]
+  X[ind == TRUE, p := p[nfeatures]]
```
martinju: Is this used to get the ranking in decent order, or something else?

R/features.R
nikolase90: I didn't want to remove it inside the function, since I want to do unit tests for the probabilities we're using. This is the same as we did before (when the Shalpey values was equal to the probabilities).

R/features.R
nikolase90: I want to keep the sampling probabilities, so that we can make sure that they are correct. This is done by unit tests in `tests/testthat/test-features.R`.

PR #142, opened 2019-11-22
Removes argument type in inv_gauss_transform

nikolase90:


nikolase90:
@martinju We'll do this one first, and the I'll reassign you as reviewer on the other [PR request](#141) once `master` is updated.

PR #147, opened 2019-11-23
Fixes notes when running R CMD check

### LINKED ISSUE #12: Fix feedback on R CMD CHECK. There should be zero errors, warnings and notes.

nikolase90:
See https://rdatatable.gitlab.io/data.table/articles/datatable-importing.html#namespace-file-namespace for more information about the problem.  

What I'm thinking with these changes:  
If we add a variable `s` in data.table inside a function we do `s <- NULL` first. If we modify or use a variable in data.table that was passed as an argument in the function, we first check if it is actually a data.table, and if so, we also check if the column we want to modify/use exists. 

**Example 1**
```
foo1 <- function() {
  x3 <- NULL
  x <- data.table::data.table(x1 = 1:10, x2 = 1:10)
  x[, x3 := 1e2]
  return(x)
}
x1 <- foo1()
x1[]
```

**Example 2**
```
foo2 <- function(x) {
  
  x3 <- NULL
  stopifnot(
    data.table::is.data.table(x), 
    !is.null(x[["x3"]])
  )
  x[, x3 := x3^2]
  return(x)
}


x2 <- foo2(foo1())
x2[]
```

Doesn't look pretty, but I'd like to avoid adding them as global variables in `R/zzz.R` as we did previously. We lose a bit a control when doing that.  

Fixes #12 .

### DIFF
--- a/R/features.R
+++ b/R/features.R
@@ -135,16 +135,14 @@ feature_not_exact <- function(m, n_combinations = 200, weight_zero_m = 10^6) {
   X[, is_duplicate := NULL]
 
   # Add shapley weight and number of combinations
-  ind <- NULL # due to NSE notes in R CMD check
   X[c(1, .N), shapley_weight := weight_zero_m]
   X[, N := 1]
-  X[between(n_features, 1, m - 1), ind := TRUE]
-  X[ind == TRUE, p := p[n_features]]
-  X[ind == TRUE, N := n[n_features]]
-  X[, ind := NULL]
+  ind <- X[, .I[between(n_features, 1, m - 1)]]
+  X[ind, p := p[n_features]]
+  X[ind, N := n[n_features]]
 
   # Set column order and key table
-  data.table::setkey(X, n_features)
+  data.table::setkeyv(X, "n_features")
   X[, id_combination := .I]
   X[, N := as.integer(N)]
   nms <- c("id_combination", "features", "n_features", "N", "shapley_weight", "p")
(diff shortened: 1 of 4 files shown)


R/observations.R
```diff
@@ -62,12 +63,12 @@ observation_impute <- function(W_kernel, S, x_train, x_test, w_threshold = .7, n
   # Remove training data with small weight
   knms <- c("index_s", "weight")
   data.table::setkeyv(dt_melt, knms)
-  dt_melt[, weight := weight / sum(weight), index_s]
+  dt_melt[, weight := weight / sum(weight), by = "index_s"]
```
martinju: Just out of curiosity: Why is this better? Is it along the lines that new version would throw an error if index_s is not in dt_melt, while the old one wouldn't if index_s was defined outside of dt_melt?

R/features.R
```diff
@@ -62,6 +62,11 @@ feature_combinations <- function(m, exact = TRUE, n_combinations = 200, weight_z
     dt <- feature_exact(m, weight_zero_m)
   } else {
     dt <- feature_not_exact(m, n_combinations, weight_zero_m)
+    stopifnot(
+      data.table::is.data.table(dt),
+      !is.null(dt[["p"]])
+    )
+    p <- NULL # due to NSE notes in R CMD check
     dt[, p := NULL]
```
martinju: I don't remember whether you answered this elsewhere, but why is this not handled within `feature_not_exact`?

R/observations.R
nikolase90: This was basically to avoid the notes. But I also checked your question, and the answer to that is yes (!). So I'll need to go over tomorrow and see if there is more cases where we should add this as well.

R/features.R
nikolase90: We have some decent unit tests on the probabilities, and I'd like to keep those for now (even though they are no longer use as the actual weights).

nikolase90:
@martin Added link which describes the problem.

PR #155, opened 2019-11-25
Corrects model type for mgcv::gam models

martinju:


nikolase90:
Nice catch! I do not why I didn't write tests for that class. We should add some tests for that as well, and see that everything lines up.

PR #157, opened 2019-11-26
Transforms to new supported models style

martinju:
Updates docs, vignette and readme. Joss paper handled separately.

### DIFF
--- a/R/models.R
+++ b/R/models.R
@@ -1,19 +1,19 @@
 #' Generate predictions for different model classes
 #'
 #' @description Performs prediction of response \code{\link[stats]{lm}}, \code{\link[stats]{glm}},
-#' \code{\link[ranger]{ranger}} and \code{\link[xgboost]{xgboost}} with binary or continuous
-#' response. See details for more information.
+#' \code{\link[ranger]{ranger}},  \code{\link[mgcv:gam]{mgcv::gam}} and \code{\link[xgboost]{xgboost}} with binary or
+#' continuous response. See details for more information.
 #'
 #' @param x Model object for the model to be explained.
 #' @param newdata A data frame (or matrix) in which to look for variables with which to predict.
 #'
 #' @details The following models are currently supported:
 #' \itemize{
-#' \item \code{\link[stats]{lm}}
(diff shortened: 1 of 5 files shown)


R/models.R
```diff
@@ -1,19 +1,19 @@
...
 #' @param newdata A data frame (or matrix) in which to look for variables with which to predict.
 #'
 #' @details The following models are currently supported:
 #' \itemize{
-#' \item \code{\link[stats]{lm}}
-#' \item \code{\link[stats]{glm}}
-#' \item \code{\link[ranger]{ranger}}
-#' \item \code{\link[mgcv]{mgcv}}
-#' \item \code{\link[xgboost]{xgboost}}
+#' \item \code{\link[stats:lm]{stats::lm}}
```
nikolase90: Why this change?

README.md
```diff
@@ -157,6 +157,11 @@ print(explanation$dt)
 
 # Finally we plot the resulting explanations
 plot(explanation)
+#> Registered S3 methods overwritten by 'ggplot2':
```
nikolase90: Regenerate the file and see if this change gets removed? I tested locally and it was removed. Let me know if I should push that.

R/models.R
martinju: Maybe not needed for lm and glm, but this is more specific and in accordance with other references to supported models in the package. In the brief description we are more "sloppy" to simplify the read.

The link text is now stats::lm instead of just lm etc, while the link is the same.

README.md
martinju: Fixed in 14a9f67

R/models.R
nikolase90: Oki. As long as the links still work, ref
> There are two tags that make it easier for people to navigate between help files
...
\code{\link[packagename]{functioname}}.
If you have a family of related functions where every function should link to every other ...

http://r-pkgs.had.co.nz/man.html

martinju: Yes, it works. See further down in that link...

\code{\link[MASS:abbey]{name}}: link to function in another package, but show name.

nikolase90: Awesome!

PR #158, opened 2019-11-26
Adds feature_labels in shapr

nikolase90:
@martinju Any input on this? 

Edited after full review:

Summarized
- Adds `features()` in `R/models.R`. Automatically extracts feature labels from model objects. Also added tests for this function in `tests/testthat/test-model.R`. 
- Adds `feature_labels` as arguments in `shapr`. Only applicable when using custom models. In that case you'll need to specify the names of the features in the model.   
- Improved error/warnings/messages/ output when calling `shapr` and `explain`.


R/shapley.R
```diff
@@ -117,17 +118,30 @@ weight_matrix <- function(X, normalize_W_weights = TRUE) {
...
 
   # Checks input argument
   if (!is.matrix(x) & !is.data.frame(x)) {
     stop("x should be a matrix or a dataframe.")
   }
 
+  # Checks input argument
+  if (is.null(feature_labels)) {
+    feature_labels <- colnames(x)
+    warning(
```
martinju:
```suggestion
    message(
```

R/shapley.R
```diff
@@ -59,6 +57,9 @@ weight_matrix <- function(X, normalize_W_weights = TRUE) {
 #' the exact method is used and all combinations are considered. The maximum number of
 #' combinations equals \code{2^ncol(x)}.
 #'
+#' @param feature_labels Character. The label of the features used for training the model. Only applicable if
```
martinju:
```suggestion
#' @param feature_labels Character vector. The labels/names of the features used for training the model.
```

R/shapley.R
```diff
@@ -59,6 +57,9 @@ weight_matrix <- function(X, normalize_W_weights = TRUE) {
...
+#' you are using a custom model.
```
martinju:
```suggestion
#' Only applicable if you are using a custom model. Otherwise the features in use are extracted from \code{model}.
```

R/utils.R
```diff
@@ -0,0 +1,8 @@
+#' @keywords internal
+unique_features <- function(x) {
```
martinju: Is this function ever used?

nikolase90: https://github.com/NorskRegnesentral/shapr/blob/6c4d6463f75a2d35488d49c0842df2724c4cfcbd/R/models.R#L296

martinju:
A few comments:

- [ ] I think we should have a warning when feature_labels are provided to a non-custom model that these are ignored and instead extracted from the model object
- [ ] I think this line https://github.com/NorskRegnesentral/shapr/blob/c34c20c78110ac6a222b65962bf45e344ef05809/R/explanation.R#L114
should be in the end of the new `explainer_x_test` function. Consider a case where the model is trained with two columns a and b, and x_test has the same columns, but in the reversed order. That will fail with the present approach, but not (I think) after the columns are reordered in `explainer_x_test`? 
- [ ] I have not tested this, and it is a corner case, but I think that if you save the xgboost model with xgb.save and load it again with xgb.load (or alternatively load an xgboost model created in e.g. the  python version of xgboost), then the object does not contain x$features, while still being of the xgb.Booster class. Not sure what is the best way to deal with that, and I am fine ignoring for now as well.
- [ ] I think we should do error handling for the feature_labels ourselves, that is, if x_test does not contain the necessary features, our package should give the appropriate error message, not the prediction function for the model. This ensures that we provide understandable error messages to the user.

R/models.R
```diff
@@ -104,7 +104,11 @@ predict_model.xgb.Booster <- function(x, newdata) {
   # Test model type
   model_type <- model_type(x)
 
-  predict(x, as.matrix(newdata))
+  if (ncol(newdata) > length(x$feature_names)) {
+    predict(x, as.matrix(newdata)[, x$feature_names, drop = FALSE])
```
martinju: I guess that always running this line (i.e. remove the if-else test would fix my issue with correct features and wrong order, but I generally think it is better than handle that outside predict_model, and rather ensure that we always provide exactly the columns that we need in the right order, to these predict_model functions.

R/models.R
```diff
@@ -302,10 +310,27 @@ features.ranger <- function(x, feature_labels) {
...
 #' @rdname features
 features.xgb.Booster <- function(x, feature_labels) {
+
+  if(!is.null(feature_labels)) feautures_message()
+
   x$feature_names
 }
+
+#' @keywords internal
+feautures_message <- function(x) {
```
martinju: Perfect, except typo in function name :-)

nikolase90:
Point number two will not be a problem for lm, glm, mgcv and ranger. Perhaps xgboost. Don't know how that package handles the names of the columns when predicting. The other packages match the name from the columns to the formula.

martinju:
> Point number two will not be a problem for lm, glm, mgcv and ranger. Perhaps xgboost. Don't know how that package handles the names of the columns when predicting. The other packages match the name from the columns to the formula.

Yes, I know. Apparently xgboost is very strict when you actually have column names, so they need to be identical (i.e. also same ordering). From the predict.xgb.Booster: 
# function (object, newdata, missing = NA, outputmargin = FALSE, 
#           ntreelimit = NULL, predleaf = FALSE, predcontrib = FALSE, 
#           approxcontrib = FALSE, reshape = FALSE, ...) 
# {
#   object <- xgb.Booster.complete(object, saveraw = FALSE)
#   if (!inherits(newdata, "xgb.DMatrix")) 
#     newdata <- xgb.DMatrix(newdata, missing = missing)
#   if (!is.null(object[["feature_names"]]) && !is.null(colnames(newdata)) && 
#       !identical(object[["feature_names"]], colnames(newdata))) 
#     stop("Feature names stored in `object` and `newdata` are different!")

martinju:
We should probably also require column names for both the xtrain and xtest?

R/models.R
nikolase90: Fixed with ea81e5e.

nikolase90:
Point two and four is fixed with d6109c9.

R/models.R
nikolase90: Should be fixed now.

R/shapley.R
```diff
@@ -75,7 +76,7 @@ weight_matrix <- function(X, normalize_W_weights = TRUE) {
 #'   \item{x_train}{data.table. Transformed \code{x} into a data.table.}
 #' }
 #'
-#' In addition to the items above \code{model} and \code{n_combinations} is also present in the
+#' In addition to the items above \code{model}, \code{feature_labels} and \code{n_combinations} is also present in the
```
martinju:
```suggestion
#' In addition to the items above \code{model}, \code{feature_labels} (updated with the names actually used by the model) 
```

R/shapley.R
```diff
@@ -75,7 +76,7 @@ weight_matrix <- function(X, normalize_W_weights = TRUE) {
...
 #' returned object.
```
martinju:
```suggestion
#' and \code{n_combinations} is also present in the returned object.
```

R/explanation.R
```diff
@@ -332,3 +336,21 @@ get_list_approaches <- function(n_features, approach) {
   }
   return(l)
 }
+
+#' @keywords internal
+explainer_x_test <- function(x_test, feature_labels) {
+
+  if (ncol(x_test) == length(feature_labels)) {
+
+    if (all.equal(colnames(x_test), feature_labels))
```
martinju:
```suggestion
    if (identical(colnames(x_test), feature_labels))
```
all.equal does not give FALSE when they are different, but rather how they are different. Identical always produce TRUE/FALSE

R/explanation.R
```diff
@@ -332,3 +336,21 @@ get_list_approaches <- function(n_features, approach) {
...
 }
+
+#' @keywords internal
+explainer_x_test <- function(x_test, feature_labels) {
+
+  if (ncol(x_test) == length(feature_labels)) {
+
+    if (identical(colnames(x_test), feature_labels))
+      return(as.matrix(x_test))
+  }
```
martinju: Suggests to delete all these lines and always go through the data.table conversion. Doesn't take much time and ensures identical $x_test output from explain (otherwise you rownames will be different)

nikolase90: Thanks, you just explained to me why the tests failed when doing the data.table conversion :-)

nikolase90: Fixed with cbc6e5e.

nikolase90:
@martinju I reverted the changes that had been made to test objects (see 9bd4c5c), and pushed that. The numerical check was OK. And then I pushed the updated in the latest commit (see d610b54). Ready for merge?

tests/testthat/test-explanation.R
```diff
@@ -103,3 +110,133 @@ test_that("Test functions in explanation.R", {
...
+        prediction_zero = p0,
+        n_samples = 1e2
+      )
+    )
+  }
+  for (i in 2:length(l)) {
+    expect_equal(l[[i - 1]], l[[i]])
+  }
+
+  # Expect silent for explainer 1, using correct, reordered and bigger data set, then identical results
```
martinju:
```suggestion
  # Expect silent for explainer 3, using correct, reordered and bigger data set, then identical results
```

nikolase90:
They changed due to the following; 
1) `feature_labels` was added to the returned object when calling `shapr`
2) The explanation object was changed due to the fact that the row names of x_test was removed when adding `explainer_x_test`.

PR #161, opened 2019-11-29
Updates .github/workflows/main.yml

nikolase90:
Some minor changes; 
- Run tests on all combinations of os and r-version 3.5/3.6. 
- Run tests when there is a PR, and when pushed into master.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,8 @@
-on: [pull_request]
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


martinju:
I mean Ubuntu 16.04

nikolase90:
Still having issues with ubuntu 18, so will only do 16.04 for now. Tests are not run when pushing to a dev branch, but they will be if you open a PR. Note that this is for Github Actions. CircleCI will run every time you push.

PR #165, opened 2019-12-06
Tests for feature labels and related warnings/messages

martinju:
Test for explanation and shapr


martinju:
@nikolase90 I don't understand why the three test that fails does so. I thought silent meant no messages, warnings or errors? When I run the functions manually there is no output to the console. Any idea?

martinju:
Hmm. devtools::test() does not fail locally with a clean environment. Don't know what the problem is now...

nikolase90:
Fails when running `devtools::check()` locally. Will see if I can figure it out.

nikolase90:
Let's just merge this one, and remove the failing things in `nikolai/feature_labels`.

PR #167, opened 2019-12-10
Fixes failing build due to incorrect cache key

nikolase90:
The cache key was incorrectly specified.


nikolase90:
@martinju Once this is merged I guess we could open the review on JOSS?

martinju:
> @martinju Once this is merged I guess we could open the review on JOSS?

Awesome!

PR #171, opened 2020-02-01
Add full MIT license plain text

frycast:
Add full MIT license to plain text file (copied from the .md file)


martinju:
Closed as @frycast approved the current license version, ref #175 and  https://github.com/openjournals/joss-reviews/issues/2027#issuecomment-581192895

PR #189, opened 2020-02-04
Updates README with installation instructions

nikolase90:
Fixes https://github.com/NorskRegnesentral/shapr/issues/185.


martinju:
@nikolase90 I suggested a slightly more concise modification. Feel free to merge if you agree.

martinju:
Also, it seems we have different versions of knitr or so, which gives differences in the bibliography depending on who compiles. Both versions are fine with me.

PR #195, opened 2020-02-05
Adds badge for JOSS-article

nikolase90:
Congrats @martinju!


nikolase90:
> Should we also add a more visible reference and link to the paper in the README and/or in the pkgdown site?

Nah. It's not that much information in the paper. If feel like it's more likely/important that person A is directed from the paper to the software, than the other way around.

martinju:
> > Should we also add a more visible reference and link to the paper in the README and/or in the pkgdown site?
> 
> Nah. It's not that much information in the paper. I feel like it's more likely/important that person A is directed from the paper to the software, than the other way around.

Mmm, sure, ok.

PR #197, opened 2020-02-06
Ctree update

martinju:
Generalized a few functions for non-3-dimensional setting.

This includes 
- writing full model matrix (identity matrix) to dt 
- generalized linear_kshap and response_mod
- bug with number of categories (levels) vs dimension
- bug using test data to train the model
- some cleaning


inst/devel_scripts/true_shapley/calculate_true_shapley_withdatatable.R
```diff
@@ -517,22 +580,20 @@ simulate_data <- function(parameters_list){
   explanation_list <- list()
   for(m in methods){
     tm_now <- Sys.time(); print(tm_now - tm_current); tm_current <- Sys.time()
-    print(paste0("Estimating Shapley value with ", m, "method"), quote = FALSE, right = FALSE)
+    print(paste0("Estimating Shapley value with ", m, " method"), quote = FALSE, right = FALSE)
     if(m == 'empirical' | m == 'empirical_ind' |  m == 'gaussian' | m == 'ctree_onehot'){
 
-      x_train_onehot <- as.matrix(x_test_onehot) #  # dt[-(1:N_testing), .(feat12, feat13, feat22, feat23, feat32, feat33)]
```
martinju: There was a bug here in the old code, probably causing results to be different with the updated code. Should update old code to verify this

martinju: Verified that fixed version of old code gives the same results as the new one

PR #203, opened 2020-05-04
Catch missing colnames

### LINKED ISSUE #202: Unclear error message when using unnamed matrix
In this example the matrices x_train and x_test will have no column names:

```data("Boston", package = "MASS")
y_var <- "medv"
ind_x_test <- 1:6
x_train <- as.matrix(Boston[-ind_x_test, x_var])
y_train <- Boston[-ind_x_test, y_var]
x_test <- as.matrix(Boston[ind_x_test, x_var])

# Later the error message will not say the matrix must have column names
colnames(x_train) <- NULL
colnames(x_test) <- NULL

model <- xgboost(
  data = x_train,
  label = y_train,
  nround = 20,
  verbose = FALSE
)
explainer <- shapr(x_train, model)
p <- mean(y_train)
explanation <- explain(
  x_test,
  approach = "empirical",
  explainer = explainer,
  prediction_zero = p
)
```
We get the following error, which makes it hard to tell the problem:

**Error in matrix(NA, ncol = n_col, nrow = no_empirical) : 
  non-numeric matrix extent**

martinju:
Fixes #202

### DIFF
--- a/R/models.R
+++ b/R/models.R
@@ -290,7 +290,7 @@ features.lm <- function(x, cnms, feature_labels = NULL) {
   if (!is.null(feature_labels)) message_features_labels()
 
   nms <- tail(all.vars(x$terms), -1)
-  if (!all(nms %in% cnms)) error_feature_labels()
+  if (!all(nms %in% cnms) | is.null(nms)) error_feature_labels()
 
   return(nms)
 }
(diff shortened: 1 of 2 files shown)


martinju:
> Could you also modify the tests in `test-models.R` so that we catch it there as well?

Did it as through `shapr`/`explain` rather than `features` to test different variants from user perspective.

martinju:
@nikolase90 Tests OK?

nikolase90:
@martinju 👍

PR #205, opened 2020-05-11
Removes circleci

nikolase90:
Fixes https://github.com/NorskRegnesentral/shapr/issues/201. 

Adds two separate workflows in `.github/workflows`
1. `lint.yaml` Linting of package.
2. `pkgdown.yaml`. Building and deploying webpage 

Should replace what we're losing when removing setup for circleci. @martinju I think that you might need to remove us from circleci. Not sure if I have still access. Will let you know.


nikolase90:
@martinju Yeah, looks like I can't remove shapr from circleci. Need to be an admin.

martinju:
> @martinju Yeah, looks like I can't remove shapr from circleci. Need to be an admin.

Ok. Think I managed to fix it. Now there is no requirement of tests passing to allow merging. I guess we should have at least one, (e.g. ubuntu-latest?). Let me know what you think.

nikolase90:
@martinju For the time being we could just have that all tests should pass to merge. The thing with circleci was that you needed to remove project shapr from circleci. So all good now.

martinju:
> @martinju For the time being we could just have that all tests should pass to merge. The thing with circleci was that you needed to remove project shapr from circleci. So all good now. 

Good! I Also had to manually remove the circle ci test from list of test required to pass otherwise it would always show as failed and merging would be blocked. The GitHub tests were never required to pass for a merge, but I could add them all.

nikolase90:
Awesome! Yeah, we should add the tests from Github actions.

PR #206, opened 2020-05-12
Inefficiency issues

### LINKED ISSUE #204: Inefficient data handling in prepare_data with exact=F
`prepare_data `stores and uses duplicates of the row containing all features if not all feature combinations are being used (`exact=F` in `shapr()`). Searches for `id_combination == 2^m` which does not exits in that case for reduction.

martinju:
3 related issues are fixed here

1.  Removed unnecessary sampling in `prepare_data `for all methods when conditioning on all or no features. For those cases we use respectively the prediction itself and the prediction_zero input, so no need to sample anything there. Should in theory reduce the runtime a little bit. Fixes #204 

2. The call to `predict_model` in `prediction` is now done only when conditioning on all or no features. This was mainly an issue before 1. was fixed, but it will still save 2 unnecessary calls. Also uses n_features to find the rows when conditioning on all or no features for consistency with 1. and improved readability.

3. When using a `n_combinations>2^m-2` in `shapr`, `feature_combinations `switches to `exact=F`, but this was not returned, so `shapr `still showed this case as exact=F. This is fixed here.

Tests updated as I added `n_features `to the output from `prepare_data()`.


R/predictions.R
```diff
@@ -48,10 +48,11 @@ prediction <- function(dt, prediction_zero, explainer) {
   stopifnot(nrow(explainer$x_test) == dt[, max(id)])
 
   # Predictions
-  dt[, p_hat := predict_model(explainer$model, newdata = .SD), .SDcols = cnms]
-  dt[id_combination == 1, p_hat := prediction_zero]
+  dt[!(n_features %in% c(0, ncol(explainer$x_test))),
+     p_hat := predict_model(explainer$model, newdata = .SD), .SDcols = cnms]
+  dt[n_features == 0, p_hat := prediction_zero]
   p_all <- predict_model(explainer$model, newdata = explainer$x_test)
-  dt[id_combination == max(id_combination), p_hat := p_all[id]]
```
nikolase90, line 54: Why isn't this ok? Shouldn't `max(id_combination)` always be equal to the case with all variables?

nikolase90: Could you give me an example so I could reproduce the problem?

R/observations.R
```diff
@@ -60,6 +62,12 @@ observation_impute <- function(W_kernel, S, x_train, x_test, w_threshold = .7, n
   )
   dt_melt[, index_s := nms_vec[id_combination]]
 
+  # Extracts zero and full conditional to be handled separately
```
nikolase90, line 65: In terms of these changes I would actually like to see that it is an improvement in speed. It think it actually might be slower. Pushing two additional rows to c++ shouldn't mean anything?

martinju:
I will try to guide you through what I have done: 

The main issue was that when using exact = F, specifics in not cause `prepare_data()` to return a two single row when conditioning on all features, but instead handled them like all other feature combinations. If the number of samples is say 500 per combinations that caused the generation of 500 extra samples from the estimated probability distribution with all features -- which are never used to do anything. Furthermore, the previous code in `prediction` then called on predict_model for all these 500 samples, just to overwrite them with p_all three lines later. 
To fix that issue, I had to introduce a `n_features `column within `prepare_data()`. Using `max(id_combinations)` or similar here, would not work when `prepare_data `is called from `explain.combined`, since `max(id_combinations)` need not corresponds to the combination with all features. 

To keep the code consistent and more transparent and easier to understand, I passed on the same convention in `prediction`. Going back to the old code, think you are right that max(id_combinations) would always correspond to conditioning on all features, and although I am sure I had produced an example where p_all was not passed correctly, I can't see how that could have happened now, so I suspect I did something wrong when saw that. So the stated reason for the change in prediction was probably wrong, but the modification of when to call `predict_model` is better, and I using n_features is consistent and easier to understand. 

Regarding the possible improvement in speed for leaving out zero and full conditioning in observation_impute it is not about two extra rows, but about `2*n_samples`. The code I wrote was just brute force, so I am sure that could be more efficient. If you really want to, I could benchmark. In any case it feels better to avoid generating a large number of rows you are just going to throw away in the next second :-) 

Also, I forgot to write in the PR text first, but when setting `n_combinations`>2^m-2 in `shapr`, `feature_combinations` switches to exact=F, but this is not returned, so the output from shapr still shows it as exact=F. This is also fixed in this PR. PR text updated.

R/observations.R
martinju: As I commented below it is about generating 2*n_samples rows, not just two. It could be changed in the c++ code, but it was much easier to do it outside.

R/predictions.R
martinju: You are right. Not relevant anymore.

nikolase90: I feel like these changes in `prediction` are unnecessary.  Here it is only 2 * number of test samples additional rows we are predicting on when doing it the old way. Searching through the entire table, using a non-key column, for those rows seems inefficient to me. The reason why we decided on using `id_combination` was that that is the column that is used throughout everything.

martinju: Yes, sure, after the changes in prepare_data, there are far fewer extra rows this is done for. I now went back to `id_combination`, but rather extracts `p_all` from `dt`, as due to the changes in `prepare_data`, x_test will always be placed in those columns and not something randomly sampled. What do you think? If you are happy, we could also remove `n_features `from the export of prepare_data as it is not needed in `prediction.`

R/observations.R
nikolase90: Sure. But the percentage these two combinations represents in this case is `2 / total number of combinations`? I don't see the incentive to introduce additional complexity and book-keeping here. This was designed this way to avoid copying the data.

martinju: Yes. The important thing is that the output has one row with the test data and weight 1 when conditioning on no features, and another copy of that when conditioning on all features. There should be no other rows with those combinations, and they should only be present if passed to `prepare_data`. So, if you think it is better to replace the rows after passing them to `observation_impute_cpp`, I could try to do that instead.

R/predictions.R
martinju: @nikolase90 What do you think about the last changes here? If you think this approach is fine, I will modify `prepare_data` not to include `n_features`.

R/observations.R
martinju: @nikolase90 What do you think about my suggestion here? Let me know if it is unclear what I mean.

nikolase90:
@martinju Give me an hour and I'll look over one more time 👍

R/observations.R
```diff
@@ -250,8 +250,9 @@ prepare_data.gaussian <- function(x, seed = 1, n_samples = 1e3, index_features =
     dt_l[[i]][, id := i]
     if (!is.null(index_features)) dt_l[[i]][, id_combination := index_features[id_combination]]
   }
+
   dt <- data.table::rbindlist(dt_l, use.names = TRUE, fill = TRUE)
-  dt <- merge(dt, x$X[, .(id_combination, n_features)], by = "id_combination")
+  dt[, n_features := x$X$n_features[id_combination]]
```
nikolase90, line 255: @martinju Avoids merging.

R/observations.R
```diff
@@ -289,7 +290,7 @@ prepare_data.copula <- function(x, x_test_gaussian = 1, seed = 1, n_samples = 1e
     if (!is.null(index_features)) dt_l[[i]][, id_combination := index_features[id_combination]]
   }
   dt <- data.table::rbindlist(dt_l, use.names = TRUE, fill = TRUE)
-  dt <- merge(dt, x$X[, .(id_combination, n_features)], by = "id_combination")
+  dt[, n_features := x$X$n_features[id_combination]]
```
nikolase90, line 293: Same here. Just to avoid merging.

nikolase90:
I just ran a benchmark on this; 
```
explanation <- explain(
    x_test,
    approach = "empirical",
    explainer = explainer,
    prediction_zero = p
  )
```
comparing against `master`. It looks like the average time spent per iteration is increasing quite a bit (from 80ms to 116ms) in the readme example. My take on this is that you should revert the changes in `observation_impute`.

martinju:
@nikolase90 Thanks for the input on this. I am happy with the result on this now after reverting some things as suggested by you and moving some other code around. 
I am just closing this as it became messy, and will open a new simpler PR tomorrow morning with the changes.

nikolase90:
@martinju Sounds good! One thing we could do is to also add a test for runtime. I.e. so that we quickly get a notification if a change will result in increased runtime.

martinju:
@nikolase90 Yes, would have been useful in this case, but we would then probably need some different use cases as some changes may give a minor increase in some cases and a bigger decrease for other cases.

PR #212, opened 2020-06-30
fix artifact cleaner

martinju:
bugfix in artifact cleaner


martinju:
Apparently there were some spacing issues caused by my text editor. Should be OK now.

PR #213, opened 2020-07-01
Improve consistency

### LINKED ISSUE #204: Inefficient data handling in prepare_data with exact=F

martinju:
3 things are handled here

1. When using a `n_combinations>2^m-2` in `shapr`, `feature_combinations `switches to `exact=F`, but this was not returned, so `shapr `still showed this case as exact=F. Fixed
2. The last line(s) in the different `prepare.data` functions were supposed to set `w=1` for when conditioning on none and all features (and reduce the number of rows for the empirical). When `exact=F`, there are fewer than `2^m` `id_combinations`, and this does not work when conditioning on all features. Therefore, this part of the code is deleted from `perpare_data`, and `predictions` are modified to remove unnecessary rows and only predict on the required rows.
3. Tests are updated according to the changes in `predictions`

This PR was based on the work in #206, also fixes #204


martinju:
A runtime comparison with the current master shows that the new version seems to be a bit faster for both `exact=F` (13%) and `exact=T` (5%). When there are more features I think speedup will be smaller, but I am confident it will not be slower.

```
library(xgboost)
library(shapr)

data("Boston", package = "MASS")

y_var <- "medv"

x_train <- as.matrix(Boston[-1:-6, x_var])
y_train <- Boston[-1:-6, y_var]
x_test <- as.matrix(Boston[1:6, x_var])

# Looking at the dependence between the features
cor(x_train)

model <- xgboost(
  data = x_train,
  label = y_train,
  nround = 20,
  verbose = FALSE
)

explainer <- shapr(x_train, model,n_combinations = 10)

p <- mean(y_train)

set.seed(123)
tt =proc.time()
for (i in 1:500){
  explanation <- explain(
    x_test,
    approach = "empirical",
    explainer = explainer,
    prediction_zero = p
  )
}
proc.time()-tt
#bb = list()
# New version
#bb[[1]] = explanation
#user  system elapsed 
#217.373   1.555  27.390 

# master
#bb[[2]] = explanation
#names(bb[[2]]$p) = 1:6
#user  system elapsed 
#312.395   3.060  31.447 

#all.equal(bb[[1]],bb[[2]])
#[1] TRUE

### EXACT ####

explainer <- shapr(x_train, model)

p <- mean(y_train)

set.seed(123)
tt =proc.time()
for (i in 1:500){
  explanation <- explain(
    x_test,
    approach = "empirical",
    explainer = explainer,
    prediction_zero = p
  )
}
proc.time()-tt
#cc = list()
# New version
#cc[[1]] = explanation
#user  system elapsed 
#280.206   2.377  34.124 

# master
#cc[[2]] = explanation
#names(cc[[2]]$p) = 1:6
#user  system elapsed 
#329.537   3.086  36.026 

#all.equal(cc[[1]],cc[[2]])
#[1] TRUE

```

martinju:
@nikolase90 Happy with this? Merge and then submit to CRAN?

PR #221, opened 2020-08-07
update manual pages for sampling

aredelmeier:
Update manual page for n_samples and sample so that there is a bit more information when choosing these for the "ctree" method.


aredelmeier:
Ya, I think you're right. I thought "sample" was for all methods but it turns out that only ctree uses it. Will update now.

PR #232, opened 2020-09-02
Fix model type bug for xgboost with categorical data and classication

martinju:
The model_type for xgboost does not return a string calling on the dummylist procedure when using classifaiction (with a probability) with xgboost (only regression). Fixing this here. Should be merged to related branches too.

### DIFF
--- a/R/models.R
+++ b/R/models.R
@@ -105,7 +105,7 @@ predict_model.xgb.Booster <- function(x, newdata) {
   # Test model type
   model_type <- model_type(x)
 
-  if (model_type == "cat_regression") {
+  if (model_type %in% c("cat_regression","cat_classification") ) {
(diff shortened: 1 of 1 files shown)


aredelmeier (commented):
Why do you have two of the same lines (ifelse(is.null(x$dummylist) etc)?

martinju:
> Why do you have two of the same lines (ifelse(is.null(x$dummylist) etc)?

What do you mean? I think they are called in different cases, depending on whether origin is true or false and giving different output? 
Should maybe rewrite the whole ifelse thing as it is not easy to read.

aredelmeier (approved):
I understand now. It wasn't clear just looking at the changes on Github. It could be simplified but this is not a priority.

PR #234, opened 2020-09-03
Fix build error on CRAN's Solaris system

martinju:
Accepted on CRAN, but fails on Solaris with the error

```
distance.cpp: In function ‘arma::cube mahalanobis_distance_cpp(Rcpp::List, arma::mat, arma::mat, arma::mat, bool)’:
distance.cpp:63:53: error: call of overloaded ‘pow(const uword&, int)’ is ambiguous
             S_scale = 1.0/pow(theseFeatures.n_elem,2);
                                                     ^
In file included from /usr/include/math.h:15:0,
                 from /opt/csw/include/c++/5.2.0/cmath:44,
                 from /home/ripley/R/Lib32/Rcpp/include/Rcpp/platform/compiler.h:100,
                 from /home/ripley/R/Lib32/Rcpp/include/Rcpp/r/headers.h:67,
                 from /home/ripley/R/Lib32/Rcpp/include/RcppCommon.h:29,
                 from /home/ripley/R/Lib32/RcppArmadillo/include/RcppArmadilloForward.h:26,
                 from /home/ripley/R/Lib32/RcppArmadillo/include/RcppArmadillo.h:31,
                 from distance.cpp:1:
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:204:21: note: candidate: long double std::pow(long double, long double)
  inline long double pow(long double __X, long double __Y) { return
                     ^
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:171:15: note: candidate: float std::pow(float, float)
  inline float pow(float __X, float __Y) { return __powf(__X, __Y); }
               ^
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:72:15: note: candidate: double std::pow(double, double)
 extern double pow __P((double, double));
               ^
*** Error code 1
make: Fatal error: Command failed for target `distance.o'
```

Inserting 2.0 for 2 in `pow` seems to do the trick. I tested the package on the Solaris system using `rhub::check_on_solaris()`
Unfortunately this fails as xgboost package is not available. but looking at the log it seems that the shapr package still installs fine after the fix (while the error above appears before the fix). I am therefore 95% sure that this fixes the issue, and think it is safe to resubmit this to CRAN.


nikolase90:
Should we also update the readme for CRAN installation? Otherwise, looks good to me!

nikolase90 (approved):

martinju:
> Should we also update the readme for CRAN installation? Otherwise, looks good to me!

Good point! I added instructions along with two CRAN badges.

martinju:
Re-running all checks now. Will submit when they are done

martinju:
v 0.1.3 accepted to CRAN and it is building without errors on all systems: https://cran.r-project.org/web/checks/check_results_shapr.html

We finally got there! Thanks for excellent work and cooperation @nikolase90 😃 

We got quite a lot of work stashed in different branches. I will over the next weeks merge them piece by piece into master moving towards a (major?) release later this autumn.

PR #239, opened 2020-11-05
Update ctree: Adding the dummy functions

martinju:
Adding the dummy functions + associated tests. Xgboost prediction code update comes in the next PR, then example on how to use it in practice in the vignette in the end.


aredelmeier (approved):
My only comment is the line: 
'# all(logical(0)) is also TRUE... be careful 
in apply_dummies()

I'm not sure if was you Martin that wrote this?

martinju:
> My only comment is the line:
> 
> # all(logical(0)) is also TRUE... be careful
> in apply_dummies()
> 
> I'm not sure if was you Martin that wrote this?

Thank you for spotting that. I checked it and it should be fine as long as dummyfunc is not modified. If for some reason someone sets fummylist$features = NULL between calling the two dummy functions, the specific line will still pass, but it will fail further down anyway so nothing to worry about.

While checking the code I found a small bug (leftover from previous version of code I think). Added a test for that + column names. 

@aredelmeier Could you check the new edits?

aredelmeier (approved):

PR #241, opened 2020-11-06
Update ctree: The rest

martinju:
Final PR with all the good stuff to update master with ctree stuff, allowing models with categorical features to be explained, estimating the conditional distrbutions using conditional inferences trees, ref https://arxiv.org/pdf/2007.01027 

Major changes
- New ctree class for explain, prepare_data, 
- Added `explainer_x_test_dt.` Same as `explainer_x_test`, but returns the data.table instead of a matrix, as needed for categorical features
- Added `ctree_comb_mincrit `class for explain for combining different mincriterion parameters for the ctree method. Analogous setup as for combining different approaches (gaussian + empirical)
- Added `create_ctree `and `sample_ctree `to fit and sample from the ctree "models".
- Readme, vignette edited accordingly and example added under `inst/script/example_ctree_method.R`
- New tests for the new functionality.

Please take a careful look, @aredelmeier 😃


R/sampling.R
```diff
@@ -146,3 +144,233 @@ sample_combinations <- function(ntrain, ntest, nsamples, joint_sampling = TRUE)
...
+#'
+#' @param x_test Matrix, data.frame or data.table with the features of the observation whose
+#' predictions ought to be explained (test data). Dimension \code{1xp} or \code{px1}.
+#'
+#' @param x_train Matrix, data.frame or data.table with training data.
+#'
+#' @param p Positive integer. The number of features.
+#'
+#' @param sample Boolean. True indicates that the method samples from the terminal node
+#' of the tree whereas False indicates that the method takes all the samples if it is less than n_samples.
```
aredelmeier, line 163: I would write: " whereas False indicates that the function takes all the observations if the number of observations is less than n_samples.

R/observations.R
```diff
@@ -241,6 +241,84 @@ prepare_data.copula <- function(x, x_test_gaussian = 1, seed = 1, n_samples = 1e
...
 
+#' @param n_samples Integer. The number of obs to sample from the leaf if \code{sample} = TRUE or if \code{sample}
+#' = FALSE but \code{n_samples} is less than the number of obs in the leaf.
+#'
+#' @param index_features List. Default is NULL but if either various methods are being used or various mincriterion are
+#' used for different numbers of conditoned features, this will be a list with the features to pass.
+#'
+#' @param  mc_cores Integer. Only for class \code{ctree} currently. The number of cores to use in paralellization of the
+#' tree building and tree prediction. Defaults to 1. Uses parallel::mclapply which relies on forking, i.e. does not
+#' work on Windows systems.
```
aredelmeier, line 252: Add something like: If working on a Windows system, mclapply() defaults to lapply()

R/sampling.R
```diff
@@ -146,3 +144,233 @@ sample_combinations <- function(ntrain, ntest, nsamples, joint_sampling = TRUE)
...
+
+    } else {
+      fit.nodes <- party::where(object = datact)
+      # newdata must be data.frame + have the same colnames as x
+      pred.nodes <- party::where(object = datact, newdata = xp)
+    }
+
+    rowno <- 1:dim(x_train)[1]
+
+    if (!sample) {
```
aredelmeier, line 232: I know we have talked about this before but this "sample" functionality seems unusually complicated for something that is not very useful (and probably only we care about it). Perhaps in another version we think about removing "sample"

README.Rmd
```diff
@@ -44,7 +44,8 @@ The following methodology/features are currently implemented:
 -   Accounting for feature dependence assuming the features are Gaussian (@aas2019explaining).
 -   Accounting for feature dependence with a Gaussian copula (Gaussian dependence structure, any marginal) (@aas2019explaining).
 -   Accounting for feature dependence using the Mahalanobis distance based empirical (conditional) distribution approach of @aas2019explaining.
--   Combine any of the three methods.
+-   Accounting for feature dependence using conditional inference trees (@hothorn2006unbiased). This methodology has currently been submitted.
```
aredelmeier, line 47: What does "This methodology has currently been submitted." mean?

inst/scripts/example_ctree_method.R
```diff
@@ -0,0 +1,40 @@
+library(shapr)
```
aredelmeier, line 1: I am surprised that this script does not include any categorical features to show how that works with shapr? Is that in another script that was submitted earlier?

tests/testthat/test-explanation.R
```diff
@@ -95,21 +95,159 @@ test_that("Test functions in explanation.R", {
...
+
+  # Ex 20: Explain predictions (ctree, sample = TRUE, default parameters)
+  ex_list[[20]] <- explain(x_test, explainer, approach = "ctree", prediction_zero = p0, sample = TRUE)
+
+  # Ex 21: Explain predictions (ctree, sample = FALSE, other ctree parameters)
+  ex_list[[21]] <- explain(x_test, explainer, approach = "ctree", prediction_zero = p0, sample = FALSE,
+                           mincriterion = 0.9, minsplit = 20, minbucket = 25)
+
+  # Ex 22: Explain predictions (ctree, sample = TRUE, other ctree parameters)
+  ex_list[[21]] <- explain(x_test, explainer, approach = "ctree", prediction_zero = p0, sample = TRUE,
```
aredelmeier, line 113: why does ex_list[[21]] exist twice (duplicate)?

aredelmeier (approved):
I have left a few comments along the way. Nothing major. Some spelling/grammar stuff for explanations and some comments for us to keep at the back of our heads for a later date.

R/sampling.R
martinju: Agree

R/observations.R
martinju: Suggesting: @param  mc_cores Integer. Only for class \code{ctree} currently. The number of cores to use in paralellization of the tree building (\code{create_ctree}) and tree sampling (\code{sample_ctree}). Defaults to 1. Note: Uses parallel::mclapply which relies on forking, i.e. uses only 1 core on Windows systems.

OK @aredelmeier ? I think it is more important to pinpoint how many cores that are used in windows than what functions it falls back to. On linux mc_cores=1 will also use lapply instead of mclapply.

R/sampling.R
martinju: @aredelmeier Please take a careful look at the new version. I think that looks better and should be doing the same. All tests are passing.

README.Rmd
martinju: Updated readme with ctree paper reference.

inst/scripts/example_ctree_method.R
martinju: There are example(s) of this in the vignette. Not use if it has ever been in a single script like this. Fixing by adding modified example from vignette.

tests/testthat/test-explanation.R
martinju: Good catch! Fixed

martinju:
> I have left a few comments along the way. Nothing major. Some spelling/grammar stuff for explanations and some comments for us to keep at the back of our heads for a later date.

Thanks for a thorough review, @aredelmeier! Could you re-review with the changes just pushed and mark the conversations accordingly?

martinju:
Looking at the ctree example once more I realize that the level checking done for categorical features is not bullet proof. While the dummy functions ensure that the training and testing data get the same levels, the original data need not have these levels registered, and that will throw an error when using then either in shapr or explain.

One suggestion is to rework the two dummyfunctions into a single function call with returns both the dummylist, the two dummymatrices and also the rafactored versions of the orignal training and testing data. 
In addition or instead of this, we should have checks in the beginning of shapr and explain that check whether the levels of the training data and testing data match those used by the model object (in xgboost via the dummylist and in lm or simlar via $xlevels) If these do not match we should throw a more informative error message. What do you think @aredelmeier ?

R/sampling.R
```diff
@@ -146,3 +144,210 @@ sample_combinations <- function(ntrain, ntest, nsamples, joint_sampling = TRUE)
...
+      fit.nodes <- party::where(object = datact)
+      # newdata must be data.frame + have the same colnames as x
+      pred.nodes <- party::where(object = datact, newdata = xp)
+    }
+
+    rowno <- 1:nrow(x_train)
+
+    use_all_obs = !sample & (length(rowno[fit.nodes == pred.nodes]) <= n_samples)
+
+    if (use_all_obs){
```
aredelmeier, line 235: Nice, I like this change!

aredelmeier: Just make sure to use "<-" and not "="

R/sampling.R
```diff
@@ -146,3 +144,210 @@ sample_combinations <- function(ntrain, ntest, nsamples, joint_sampling = TRUE)
...
+                        control = partykit::ctree_control(minbucket = minbucket,
+                                                          mincriterion = mincriterion,
+                                                          splitstat = "maximum"))
+      })
+    } else if (use_partykit == "never") {
+      datact <- party::ctree(fmla,
+                             data = df,
+                             controls = party::ctree_control(minbucket = minbucket,
+                                                             mincriterion = mincriterion))
+    } else {
```
aredelmeier, line 343: This is probably not important, but you don't check that use_partykit is "never", "always" or "on_error" so the else statement could be anything (not just for "always".

README.md
```diff
@@ -47,23 +47,25 @@ This package implements the methodology of Aas, Jullum, and Løland
...
+    Løland (2019)).
+-   Accounting for feature dependence using the Mahalanobis distance
     based empirical (conditional) distribution approach of Aas, Jullum,
     and Løland (2019).
-  - Combine any of the three methods.
-  - Optional use of the AICc criterion of Hurvich, Simonoff, and Tsai
-    (1998) when optimizing the bandwidth parameter in the empirical
+-   Accounting for feature dependence using conditional inference trees
+    (Redelmeier, Jullum, and Aas (2020)).
+-   Combination any of the four methods.
```
aredelmeier, line 63: A combination of any of the four methods or
Combining any of the four methods.

README.md
```diff
@@ -75,10 +77,10 @@ Current methodological restrictions:
 
 Future releases will include:
 
-  - Support for parallelization over explanations, Monte Carlo sampling
+-   Support for parallelization over explanations, Monte Carlo sampling
     and features subsets for non-parallelizable prediction functions.
-  - Computational improvement of the AICc optimization approach
-  - Adaptive selection of method to account for the feature dependence
+-   Computational improvement of the AICc optimization approach
```
aredelmeier, line 82: Period at the end of the sentence. And for the next sentence.

README.md
```diff
@@ -75,10 +77,10 @@ Current methodological restrictions:
...
+-   Adaptive selection of method to account for the feature dependence
 
 Note that both the features and the prediction must be numeric. The
 approach is constructed for continuous features. Discrete features may
```
aredelmeier, line 86: may what?

inst/scripts/example_ctree_method.R
aredelmeier: You're right, I can't find an example.  But thanks.

R/observations.R
aredelmeier: Great.

README.Rmd
aredelmeier: Great.

R/sampling.R
aredelmeier: This looks good to me.

aredelmeier (approved):
Let's discuss your issues with dummy_function on Teams?

inst/scripts/example_ctree_method.R
martinju: For some reason that part of the code does not show up when showing all commits under "files changes". You should see it if you select "changes from last review" in the upper when in the "file changes" view.

R/sampling.R
martinju: Ok now?

aredelmeier: I don't see the difference?

README.md
martinju: It's on the next line. You need to click the arrow-thing on the left side (right by the line numbers) to see more lines.

R/sampling.R
martinju: Select "show all commits" or "commits since last time you reviewed" in the upper left corner under the file changes view.

martinju:
> Looking at the ctree example once more I realize that the level checking done for categorical features is not bullet proof. While the dummy functions ensure that the training and testing data get the same levels, the original data need not have these levels registered, and that will throw an error when using then either in shapr or explain.
> 
> One suggestion is to rework the two dummyfunctions into a single function call with returns both the dummylist, the two dummymatrices and also the rafactored versions of the orignal training and testing data.
> In addition or instead of this, we should have checks in the beginning of shapr and explain that check whether the levels of the training data and testing data match those used by the model object (in xgboost via the dummylist and in lm or simlar via $xlevels) If these do not match we should throw a more informative error message. What do you think @aredelmeier ?

Will handle this in another PR.

R/sampling.R
martinju: @aredelmeier Let me know if you still can't see it and/or you have found it and it is ready for merge.

aredelmeier: Great. Now I see it.

aredelmeier (approved):

PR #243, opened 2020-11-12
Merge make dummies

aredelmeier:


R/features.R
```diff
@@ -176,53 +176,95 @@ helper_feature <- function(m, feature_sample) {
...
-#' x_test <- Boston[1:6, x_var]
+#' # convert to factors for illustational purpose
+#' x_train$rm <- factor(round(x_train$rm))
+#' x_test$rm <- factor(round(x_test$rm), levels = levels(x_train$rm))
 #'
-#' dummylist <- make_dummies(data = rbind(x_train, x_test))
+#' dummylist <- make_dummies(data = rbind(x_train, x_test), newdata = x_test)
 #'
-make_dummies <- function(data) {
+make_dummies <- function(data, newdata) {
```
martinju, line 214: Suggesting data -> x_train or data_train, newdata -> x_test or data_test. There is nothing new about newdata here (while there actually is in `apply_dummies`.

R/features.R
```diff
@@ -176,53 +176,95 @@ helper_feature <- function(m, feature_sample) {
...
   }
-  data <- data.table::as.data.table(as.data.frame(data, stringsAsFactors = FALSE))
+  if (is.null(colnames(newdata))) {
+    stop("newdata must have column names.")
+  }
+  if (is.null(newdata)) {
+    stop("newdata needs to be included.")
+  }
 
+  data <- data.table::as.data.table(as.data.frame(data, stringsAsFactors = FALSE))
```
martinju, line 228: For this and similar places: 

We say that data and newdata must be data.table or data.frame, but still we call as.data.frame, so it could have been matrix. Do we need to handle matrix? If not, we could skip the cumbersome as.data.frame before as.data.table here. We could/should then check that there are no other column classes than is.numeric and is.factor. 

If one passes a data.frame with characters, currently this converts the data.frame characters to a factor with the levels present in that data.frame. If the same or not found in the test data, then the error of not identical levels might be confusing as the user have not specified any levels, our function has created it for them. If we require factor from the beginning we avoid this.

R/features.R
```diff
@@ -176,53 +176,95 @@ helper_feature <- function(m, feature_sample) {
...
     stop("Features must have unique names.")
   }
-  p <- sapply(data[, features, with = FALSE], is.factor)
-  p_sum <- sum(p)
 
-  if (p_sum > 0) {
-    factor_features <- features[p]
+  # Check that all features in data are in newdata
+  if (!all(features %in% names(newdata))) {
+    stop("Some features missing from newdata.")
```
martinju, line 239: I think we should specifiy what features are missing doing something like 
paste0("The feature(s) ",paste0(features[!(features %in% names(newdata))],collapse = ", ")," are missing from newdata.")

martinju (commented):
Have you considered combining the training data and testing data in make_dummies, extract the factors, contrasts etc, and then just calling apply_dummies both for the training and test data? I would imagine that would make the code cleaner, but there might be issues that I am not thinking of. 

If that is not feasible I would change some variable names in make_dummies like is_factor_all, which refers to data, and not both data and newdata, while is_factor_new refers to the new data (which is the testing data). Using the train and test tags instead of data and newdata makes it easier for the user to understand the error messages too (they just passed the training and testing data).

R/features.R
aredelmeier: Thanks for the suggestion. I have now done this.

R/features.R
```diff
@@ -176,53 +176,95 @@ helper_feature <- function(m, feature_sample) {
 #'
 #' @param data data.table or data.frame. Includes all the features (both factors and possibly others).
 #'
+#' @param newdata data.table or data.frame. New data (features) that has the same
+#' features as \code{data}.
+#'
 #' @return A list that contains the following entries:
 #' \describe{
+#' \item{obj}{List, Contains \describe{
 #' \item{features}{Vector. Contains the names of all the features in \code{data}.}
```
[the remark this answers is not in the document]
aredelmeier: Good point! Yes "data" is needed when using apply_dummies() in shapr()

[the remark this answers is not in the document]
martinju: Are you sure? It seems all you use obj$data for within apply_dummies you have already stored in other parts of obj.

R/features.R
aredelmeier: This is dealt with.

R/features.R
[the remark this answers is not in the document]
aredelmeier: Good point! Fixed

R/features.R
```diff
@@ -191,117 +203,177 @@ helper_feature <- function(m, feature_sample) {
...
-#' y_train <- Boston[-1:-6, y_var]
-#' x_test <- Boston[1:6, x_var]
+#' dummylist <- make_dummies(traindata = x_train, testdata = x_test)
 #'
-#' dummylist <- make_dummies(data = rbind(x_train, x_test))
-#'
-make_dummies <- function(data) {
+make_dummies <- function(traindata, testdata) {
+
+  contrasts <- features <- factor_features <- model.frame <- model.matrix <- NULL # due to NSE notes in R CMD check
```
martinju, line 220: I have not tried, but is model.frame <- model.matrix <- NULL really needed? They are just functions...

R/features.R
```diff
@@ -191,117 +203,177 @@ helper_feature <- function(m, feature_sample) {
...
-  data <- data.table::as.data.table(as.data.frame(data, stringsAsFactors = FALSE))
 
+  if (!all(sort(colnames(traindata)) == sort(colnames(testdata)))) {
+    stop("traindata and testdata must have the same column names.")
+  }
 
-  features <- colnames(data)
+  features <- colnames(traindata)
+  # feature names must be unique
   if (length(unique(features)) < length(features)) {
```
martinju, line 241:
```suggestion
  if (any(duplicated(features))) {
```

R/features.R
```diff
@@ -191,117 +203,177 @@ helper_feature <- function(m, feature_sample) {
...
-    factor_list <- lapply(data[, factor_features, with = FALSE], levels)
+  # In case the testing data has a different column order than the training data:
+  testdata <- testdata[, features, with = FALSE]
 
+  # Check if the features all have class "numeric" or "factor
+  if (!all(sapply(traindata, class) %in% c("numeric", "factor"))) {
+    stop("All traindata must have class numeric or factor.")
+  }
+  if (!all(sapply(testdata, class) %in% c("numeric", "factor"))) {
+    stop("All testdata must have class numeric or factor.")
```
martinju, line 258:
```suggestion
  # Check if the features all have class "integer", "numeric" or "factor
  if (!all(sapply(traindata, class) %in% c("integer", "numeric", "factor"))) {
    stop("All traindata must have class integer, numeric or factor.")
  }
  if (!all(sapply(testdata, class) %in% c("integer", "numeric", "factor"))) {
    stop("All testdata must have class integer, numeric or factor.")
```

R/features.R
```diff
@@ -191,117 +203,177 @@ helper_feature <- function(m, feature_sample) {
...
-#'
-#' dummylist <- make_dummies(data = rbind(x_train, x_test))
-#'
-#' x_train_dummies <- apply_dummies(obj = dummylist, newdata = x_train)
+#' @keywords internal
 #'
-apply_dummies <- function(obj, newdata) {
+apply_dummies <- function(obj, testdata) {
 
   features <- model.frame <- model.matrix <- NULL # due to NSE notes in R CMD check
```
martinju, line 337: Same as above. model.frame and model.matrix should not be needed, I think.

R/features.R
```diff
@@ -191,117 +203,177 @@ helper_feature <- function(m, feature_sample) {
...
-  newdata <- data.table::as.data.table(as.data.frame(newdata, stringsAsFactors = FALSE))
 
+  # in case the testing data has a different column order or more columns than the training data:
+  testdata <- testdata[, features, with = FALSE]
 
-  # check all features are in newdata
-  if (!all(obj$features %in% names(newdata))) {
-    stop("Some features missing from newdata.")
+  if (!all(sapply(testdata, class) %in% c("numeric", "factor"))) {
+    stop("All testdata must have class numeric or factor.")
```
martinju, line 356:
```suggestion
  if (!all(sapply(testdata, class) %in% c("integer", "numeric", "factor"))) {
    stop("All testdata must have class integer, numeric or factor.")
```

martinju (commented):
See the few minor comments. The rest looks good. I will wait for the tests to approve.

inst/scripts/example_ctree_method.R
```diff
@@ -64,12 +82,17 @@ model_cat <- xgboost::xgboost(
   nround = 20,
   verbose = FALSE
 )
-model_cat$dummylist <- dummylist
+model_cat$dummylist <- make_dummies_list$obj
+
+explainer_cat <- shapr(make_dummies_list$traindata_new, model_cat)
 
-explainer_cat <- shapr(x_train_cat, model_cat)
+# Spedifying the phi_0, i.e. the expected prediction without any features
```
martinju, line 89:
```suggestion
# Specifying the phi_0, i.e. the expected prediction without any features
```

tests/testthat/test-features.R
```diff
@@ -170,193 +170,205 @@ test_that("Test helper_feature", {
...
+  x_test$rm <- factor(round(x_test$rm), levels = levels(x_train$rm))
 
   numeric_feat <- !sapply(x_train, is.factor)
   nb_numeric_feat <- sum(numeric_feat)
 
-  dummylist <- make_dummies(data = rbind(x_train, x_test))
-  x_train_dummies <- apply_dummies(obj = dummylist, newdata = x_train)
+  dummylist <- make_dummies(traindata = x_train, testdata = x_test)
 
+  # have to call devtools::load_all() for internal functions
```
martinju, line 315:
```suggestion
```

martinju (commented):
Just tiny changes. Will commit them + some documentation stuff done locally before merging.

martinju (commented):
tiny fixes

martinju (approved):

aredelmeier (commented):
This looks fine to me

PR #247, opened 2021-01-21
Update R-CMD-check.yaml

martinju:
try forcing macos-10.15


martinju:
The macOs GHA issue got magically solved after linting and styling, while doing nothing else. Closing.

PR #256, opened 2021-02-02
Add categorical method

aredelmeier:
This branch includes the "categorical" method that can be used to estimate the Shapley values when the model includes only categorical features. This pull request is a draft since we don't actually want to merge into master yet. 

The categorical method can be used either with a ready "joint_prob_dt" object which contains all joint probabilities or can estimate the probabilities based on the training data. 

The main work is in:  explanation.R (see explain.categorical), observations.R (see prepare_data.categorical), and the tests in test-explanation.R (see "test functions in explanation.R with factor features"). 

I also had to add model_objects (lm_model_with_cat_object.rds) and test_objects (explanation_explain_cat_obj_list_fixed.rds, explanation_explain_cat_obj_list.rds, shapley_explainer_cat_obj.rds). 

I started making tests for models that combine both numerical and categorical data but since this was based only on the ctree method, I removed these tests so we can create a new PR when finished. These tests can be found in a commit on Feb 2, 2021.


martinju:
This is handled in #307 instead

PR #264, opened 2021-06-29
Update README.md

andersloland:


martinju (changes requested):
Good idea, but don' modify the md file directly. Instead, update the bib file under inst used by the rmd file. That should regenerate the md file.

PR #267, opened 2021-08-10
Improve user experience

### LINKED ISSUE #222: Move n_samples argument to explain
Currently, the `n_samples` argument is "hidden" in `prepare_data`, being passed on from `explain`. Since this is a fairly useful parameter, it should be exposed as a parameter in `explain`, just like the similar parameter `sample` to be easier to spot for users.

### LINKED ISSUE #257: Improve documentation for w_threshold
The `w_threshold `parameter for the empirical method is poorly documented. It is the \eta-parameter in https://arxiv.org/abs/1903.10464 
It is currently only implicitly mentioned under the comparison to Lundberg and Lee's implementation in the vignette.

### LINKED ISSUE #259: Add prefix Note/Warning to console print
We print a series of "messages" to the console when e.g. a model does not store feature label classes. Either switch to using warning() or prefix the message with something like "Success with note: ..." to clarify to the reader that it completed fine, but that the content of the message should be noted".

martinju:
- Move `w_threshold` to `explain` (fix #257)
- Move `n_samples` to `explain` (fix #222)
- Add "Success with message:" to all messages to make it clear they are not errors or warnings. (fix #259 )
- Add reference to paper and some details in `explain` documentation

(+ fix linting script on GHA)


R/explanation.R
```diff
@@ -19,7 +22,6 @@
 #' @details The most important thing to notice is that \code{shapr} has implemented four different
 #' approaches for estimating the conditional distributions of the data, namely \code{"empirical"},
 #' \code{"gaussian"}, \code{"copula"} and \code{"ctree"}.
-#'
 #' In addition, the user also has the option of combining the four approaches.
 #' E.g. if you're in a situation where you have trained a model the consists of 10 features,
```
aredelmeier, line 26:
```suggestion
#' E.g., if you're in a situation where you have trained a model that consists of 10 features,
```

R/explanation.R
```diff
@@ -29,6 +31,13 @@
 #' \code{"approach[i]" = "gaussian"} it means that you'd like to use the \code{"gaussian"} approach
```
aredelmeier, line 31:
```suggestion
#' \code{"approach[i]" = "gaussian"} means that you'd like to use the \code{"gaussian"} approach
```

aredelmeier (approved):
Looks good to me.

PR #268, opened 2021-08-10
Update GHA

### LINKED ISSUE #214: Switch to aliases for R versions in GitHub actions
No rush at all, but maintaining testing as R gets updated seems to be easier using aliases instead of hardcoded R versions in the GitHub actions setup.

See https://github.com/r-lib/actions/commit/c2687dac11bca64e304f295084aa58050d90811d and  https://github.com/r-lib/actions/issues/26

What do you think @nikolase90 ?

martinju:
The current Github Actions R-CMD-checks are outdated.

- Replacing R-CMD check script by this
https://github.com/r-lib/actions/blob/master/examples/check-full.yaml
which updates the ubuntu version and switches to aliases to also fix #214 
- Also added duplicate checks which for older R-versions (oldrelease + 3.6) which does not build the vignette (as that has a tendency to cause issues with stringi etc)

### DIFF
--- a/.github/workflows/R-CMD-check-no-vignette.yaml
+++ b/.github/workflows/R-CMD-check-no-vignette.yaml
@@ -23,10 +23,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: 'release'}
-          - {os: macOS-latest, r: 'release'}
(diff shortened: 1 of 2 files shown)


aredelmeier (approved):
I am not an expert with the different operating systems and R versions but this looks fine to me.

PR #269, opened 2021-08-17
Bugfix empirical method with independence method

martinju:
I discovered a bug in the implementation of the independence method (type == "independence" and approach = "empirical") which gave biased results in the case where just a portion of the full training data set was used for the Monte Carlo integration AND one or more of the features were discrete/categorical (i.e. took just a few unique values).

The issue was that rows in the training set were not randomized correctly for subsets containing features with only a few unique values. Adding just a bit of noise to the (hacked) distance matrix made the trick. 

The script under /inst/scripts/devel/indep_comp_test.R illustrated the problem (to be deleted upon merge).

Note1: This did not affect the case where the full data set is used for the Monte Carlo integration (which is the default whenever there are less than 1000 training samples). 

Note2: The independence method will be implemented very soon, to then replace this implementation.


aredelmeier (approved):
I didn't go through the script

PR #271, opened 2021-08-24
Update R-CMD-check.yaml

martinju:
Test setting repo


martinju:
Just a test, this was not the issue (it was apparently adding .Rprofile with some python stuff)

PR #273, opened 2021-08-25
Causal and asymmetric Shapley values implementation

igbucur:
This branch contains an implementation for computing causal and asymmetric Shapley values, based on the supplement code for the paper [1]. The code is adapted from the package CauSHAPley (https://gitlab.science.ru.nl/gbucur/caushapley/).

_Asymmetric Shapley values_ were proposed in [2] as a way to incorporate causal knowledge in the real world by restricting the possible permutations of the features when computing the Shapley values to those consistent with a (partial) causal ordering.
_Causal Shapley values_ were proposed in [1] as a way to explain the total effect of features on the prediction, taking into account their causal relationships, by adapting the sampling procedure in shapr. 
The two ideas can be combined to obtain _asymmetric causal Shapley values_. For more details, see [1].

The branch adds the following functions for computing causal Shapley values:
- _sample_causal_ in _sampling.R_
- _explain.causal_ in _explanation.R_
- _prepare_data.causal_ in _observations.R_

The branch adds the following functionality for computing asymmetric Shapley values:
- additional branches for _feature_combinations_ and _feature_exact_ in _features.R_
- _respects_order_ in _utils.R_

Finally, the function _shapr_ gets two new parameters:
- _asymmetric_ : Logical flag specifying whether we want to compute asymmetric Shapley values. 
- _causal_ordering_ : List of vectors specifying (partial) causal ordering.

These parameters are saved in the explainer object returned by _shapr_, for which reason the known objects in the test suite have been updated. The branch also adds a number of basic tests for the new functionality.

**References:**
[1] Heskes, T., Sijben, E., Bucur, I. G., & Claassen, T. (2020). Causal Shapley Values: Exploiting Causal Knowledge to Explain Individual Predictions of Complex Models. Advances in Neural Information Processing Systems, 33.
[2] Frye, C., Rowat, C., & Feige, I. (2020). Asymmetric Shapley values: incorporating causal knowledge into model-agnostic explainability. Advances in Neural Information Processing Systems, 33.


martinju:
@igbucur Are you currently working on this? If so, let me know if you want to chat about how to go about it!

igbucur:
@martinju Yes, I think I'm ready to give it a go. I had a look at how to tackle the proposed extension and here are my thoughts:

1. For the _causal_ part, the solution is as you hinted before. I think the function you suggested would have to replace `lapply` in each of the `prepare_data` functions. The function would take as input the causal ordering and sample each different causal chain component , which means it would call the available sampling functions multiple times (these would not have to be changed). In the default case, there is a single chain component containing all variables, so the new function would call the sampling functions just once for all the variables and for all possible conditioning sets, as is being done at the moment.

2. For the _asymmetric_ part, it should be even simpler. The features that do not follow the causal ordering have to be removed if `asymmetric == TRUE`. We could do that by passing an appropriate `index_features` argument to the `prepare_data` functions. This could probably be done in the `explain` function.

What do you think? Does this approach seem reasonable?

igbucur:
Okay, I will think how it could be done in the `shapr` function. The difficulty here I think is that instead of calling the sampling functions just once for each feature, you have to call them multiple times per feature, that is, the number of components times.

I was thinking about encapsulating the causal ordering functionality either in a new custom lapply or somewhere upstream (perhaps in `shapr`, like you suggested) in order to avoid having to reimplement this factorization every time a new sampling functions is added. This way the sampling functions can stay the same, while the decision on which conditional probabilities need to be estimated and multiplied is made upstream. It might also be an idea to design a function that takes the features in and the causal ordering and splits them according to the different components. Perhaps this is something that could be done in `shapr`.

martinju:
And let me know if you want me to out together a function like that is shapr. It should be rather straightforward, I think.

LHBO:
Closed the PR, as this has been implemented as part of shapr version 1.0.0 in #402 through PR #400. There, the asymmetric and causal methodologies were extended to all applicable approaches in the shapr package.

PR #275, opened 2021-08-31
Jens/explain batch

JensWahl:
This PR adds the option to estimate Shapley values in batches by looping over the `S` matrix. 

Changes: 

* New argument `n_batches` in `explain` specifying the number of batches. 
* `set.seed` is moved from `prepare_data` to `explain`. 
* The argument `only_return_contrib_dt` is a new argument that can be passed to explain through `...` and makes `explain` return the contribution matrix for all test observations. This is used by `explain.combined` and `explain.ctree_comb_mincrit`. 
* `prediction` only returns `p` and `dt_mat`. The rest is moved to `prepare_and_predict` which now return the `shapr` object. 
* New function `compute_shapley` that computes Shapley values.  

The method gives equal results when `n_batches = 1` for all methods except `explain.combined` as `set.seed` is moved from  `prepare_data` to `explain`. This leads to setting the seed only once, instead of setting the same seed again for each `approach` passed to `explain.combined`. This is why the tests fail for `ex_list[10, 14, 30, 31, 37]` in `test-explanation.R` as they include multiple methods that do sampling, i.e. `gaussian, copula` and `ctree`. 

Other changes not directly related

* Allow up to 30 groups (the same as the number of features)


R/explanation.R
```diff
@@ -603,7 +603,7 @@ create_S_batch <- function(explainer, n_batches, index_features = NULL) {
 
   if (!is.null(index_features)) {
     # Rescale the number of batches to the percentage of observations used
-    n_batches <- floor(length(index_features) / nrow(explainer$S) * n_batches)
+    n_batches <- max(1,floor(length(index_features) / nrow(explainer$S) * n_batches))
```
martinju, line 606: @JensWahl Please ensure this fix does not break anything else. This caused a problem as it returned n_batches = 0, not 1 when used with a combination approach.

R/explanation.R
```diff
@@ -444,7 +444,12 @@ explain.combined <- function(x, explainer, approach, prediction_zero, n_samples
                          only_return_dt_mat = TRUE, seed = NULL, ...)
   }
 
+  dt_l_keep <<- dt_l
+
   dt_mat <- unique(rbindlist(dt_l))
+  setkey(dt_mat, row_id)
```
martinju, line 448: @Jens The main issue was that setkey was not called on the entire dt_mat (but only on the sub-dt_mat's), such that the ordering of the rows was incorrect.

R/explanation.R
JensWahl: Thanks, I actually fixed this issue on a local test branch, but didn't push ut.

martinju:
@JensWahl Before mergning, it would be great if you could be more detailed on what functional changes has been made in the PR description, altså those not directly relevant to this PR (see the latest PRs) + remember to dump the version number.

martinju (approved):

martinju:
Note: We should update vignettes and maybe the readme with the changes made here later on (and since the last CRAN release)

PR #277, opened 2021-09-17
Adding non-exact grouping

martinju:
This PR adds non-exact grouping, i.e. the kernelSHAP method of sampling n_combinations of the 2^no_groups  different groups available to use in the computation. It is essentially a duplicate of the code for features.

Other edits:
- Restructured the if-else checks for whether to switch to exact when a too large number of n_combinations is specified. 
- switched to update=F for one test against a disk stored file


JensWahl (approved):
Looks good to me.

PR #279, opened 2021-09-17
Allow passing of covariance matrix from explain to empirical

martinju:
For the Gaussian method it is possible to pass the covariance matrix (and mean) of the features from explain(). This PR allows the same for the empirical method (which uses it in the computation of the Mahlanobis distance). 

If not provided the covariance matrix is estimated by the empirical covariance matrix of the (training) data (from the explainer-object). This is now done in the explain object as for the Gaussian method instead of within the distance function in prepare_data.

- Documentation updated
- Tests for equality when passing the covariance matrix and computing it internally are added
- The version number is dumped

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: shapr
-Version: 0.2.0.9000
+Version: 0.2.0.9001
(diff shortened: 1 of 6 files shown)


JensWahl (approved):
Looks good!

PR #283, opened 2021-11-03
Bugfix prepare_data.independence

martinju:
`prepare_data.independence` used to return the incorrect `id_combinations`. This caused an issue when using `n_batches>1`, but not for `n_batches=1`.
This is now fixed and test added.


JensWahl (approved):

JensWahl:
LGTM!

PR #291, opened 2022-07-06
Newplots continued

ungvilde:


martinju:
TODO @ungvilde 

- Update docs with beeswarm + scatter in plot_type
- Update docs with default colors (I changed the default input to NULL in which it uses the hardcoded defaults we used before per plot_type). Make it clear what colors are used as defaults per method in the docs
- Move all blabla <- NULL to the top of the functions

martinju:
We'll merge this now as it works, but it does currently not handle categorical features form beeswarm, and probably not for scatter either. Will fix that in another PR.

PR #292, opened 2022-07-06
Jens/airquaility data

JensWahl:
* Replace the `Boston` dataset from `MASS` with the `airquality` dataset from `datasets`.
* Update the `README` and the vignette with the new API.


martinju (commented):
Readme and vignette looks good now.
For explain() I removed the entire example earlier. Can you add it back with the updated style/data? 
Pick it from https://github.com/NorskRegnesentral/shapr/blob/master/R/explanation.R

martinju (approved):

PR #293, opened 2022-07-07
Progress bar

ungvilde:
`progressor ` has been added to the `run_batch`  function in `compute_vS`. 

The progress bar counts how many of the batches have been completed. Even though the batches may have non-equal numbers of elements, and therefore make up different proportions of the total computing time, each batch computation is considered an equally large step of the total computation.

The progress bar will be more accurate if we specify the proportion each batch make of the total computation. See https://github.com/HenrikBengtsson/progressr/issues/13 for how it might be handled in the future.


martinju:
I tried to modify this a bit to make it more user-friendly, but did not succeed. I tried to move the set default handlers if none are provided, but the defaults are not obeyed for some reason. 

Also, the vignette does not seem to like the handlers(global=TRUE). We could disable it in the vignette, but that requires an extra argument for every call to explain() so not sure its a good idea.

@ungvilde Can you take look at my attempt and see if you can sort it out? See the check_progress.R example script for examples.

PR #303, opened 2022-08-30
Jens/fix interactions

JensWahl:
Make it possible to have lm/glm/gams models with interactions.


martinju (changes requested):
Can you add an output test which uses interactions? 
Add such a model to `helper-lm.R` under test/testthat, then add a single snapshot-test with interactions in `test-output.R`

martinju (approved):

PR #304, opened 2022-08-30
Jens/order model and train

JensWahl:


martinju:
@JensWahl Looks good. Can you add a test for this in test-setup? I suggest a setup where you have 2 lm-models, which were trained with the same data, but the columns where in different orders so the labels provided by get_model_specs are in different order. Running both models through explain should work with success (test this) and their output should be identical (test this). Please re-use one of the existing lm-models from the helper-lm file.

PR #307, opened 2022-09-06
Annabelle/cat approach

aredelmeier:
Hi Martin,
 
Would be great if you took at the categorical approach to make sure it looks like the other approaches.

One question I had was with passing an argument to explain. Should this be accessed later through internal$parameters or some other way? (Couldn't find the answer looking through the Gaussian method). 

No tests yet but I manually checked that the results give the same output as the old shapr package. 

No need to merge the branch until the tests are done. 

Annabelle


R/approach_gaussian.R
```diff
@@ -14,10 +14,14 @@
 setup_approach.gaussian <- function(internal,
                                     gaussian.mu = NULL,
                                     gaussian.cov_mat = NULL, ...) {
+
+
   parameters <- internal$parameters
   x_train <- internal$data$x_train
   feature_specs <- internal$objects$feature_specs
 
+  # TO DO: gaussian.mu should probably be extracted from internal$parameters$gaussian.mu...
```
martinju, line 23: Yes, you are right. This is a bug. I suggest doing it through the insert_defaults function as for ctree. (For gaussian the defaults are NULL, so it is not strictly necessary, but consistency is nice). Should not fix that in this PR, though.

martinju (approved):
I made some minor changes, prefixing the input parameters by categorical, inserting default values and move some documentation, and adding NULL to fix NSE notes on checks. I have not checked the code carefully, but I do think that was decent already 2 years ago.

R/approach_gaussian.R
aredelmeier: I'll make another PR with this.

PR #310, opened 2022-09-09
Jens/fix plot

JensWahl:
This is a draft. 

I moved all the ordering of the data into its own function and changed the names used in the function to be more consistent with the rest of the package. I also suppressed the warning when `plot_type = "beeswarm"`. 

I also added a snapshot test for `n_combinations`.


martinju (changes requested):
This looks good, but is there a way to get around the warnings without suppresswarning?

Also, can you make scatter and beeswarm work with categorical features?

martinju (approved):

martinju:
@JensWahl wasn't quite done 🙂
Some checks still fails. Will hopefully fix the remaining once tomorrow.

PR #314, opened 2022-09-12
Annabelle/timeseries approach

aredelmeier:
Adding time series method.

Example script to run is called ```inst/time_series_annabelle.R``` which also shows how to simulate somewhat realistic time series data.

This PR implements the method from the ```groupShapley_paper_experiments``` branch and now returns the same Shapley values as this branch. 

This approach takes in two parameters: ```timeseries.fixed_sigma_vec``` (same as empirical approach) and ```timeseries.bounds``` (optional) upper and lower bounds to foor/ceiling the sampled time series data.  

I apologize for the terrible commit messages but I was going back and forth between the two branches to try to get the same Shapley values.


martinju (approved):

martinju:
I think we can just go ahead and merge this.

PR #319, opened 2022-11-27
conditional skip for unavailable partykit

MichaelChirico:


martinju:
Same as #320.  Thanks in any case :-)

PR #321, opened 2022-11-29
Expose an option to run the procedure sequentially in explain.

jonlachmann:
For debugging purposes it seemed beneficial to be able to use the non-parallel version of the code. It is also useful if many parallel explanations are run, to avoid over-subscription of the available cores (i.e. running more threads than cores, causing slowdowns).

### DIFF
--- a/R/compute_vS.R
+++ b/R/compute_vS.R
@@ -4,14 +4,14 @@
 #' @inheritParams default_doc
 #' @inheritParams explain
 #'
-#' @param method Character
+#' @param parallel Logical.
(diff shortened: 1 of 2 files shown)


martinju:
Thanks! I agree this is useful for devel purposes (which is the reason for putting it in there in the first place). However, I am unsure whether it is wise to add another argument to `explain()`, as there are already quite many. I will leave this open for now, while thinking it through.

martinju:
Closing for now as only useful for development (and there are already many arguments in `explain()`)

PR #328, opened 2023-01-24
Add a function for explaining forecast models, explain_forecast.

jonlachmann:


martinju (commented):
Just crossing out the files that are fully OK here.

martinju:
TODOs:

- [ ] Document explain_forecast
- [ ] Update vignette with usage example for explain_forecast()
- [ ] Consider modifying setup to work with forecast instead of creating a very duplicate function (MJ will look at this)
- [ ] Handle failing checks (MJ)
- [ ] Go through and approve changes in other functions (MJ)

martinju:
After merging another branch into devel I did some updating to make tests and checks pass again. Nothing major related to what we do in this PR.

PR #345, opened 2023-06-16
Added keras support

rawanmahdi:
Implemented keras predict function. I also added an example using the shapr iris dataset on a sequential keras model and running explain(). I have not yet figured out how to implement the get_model_specs function for keras, as I am not sure if there is a way to access feature names through the model alone. This is my first PR so apologies in advance for any issues, and please let me know if theres anything to add/fix! Thanks :)

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -54,7 +54,7 @@ df_shapley, pred_explain, internal, timing = explain(
 print(df_shapley)
 ```
 
-`shaprpy` knows how to explain predictions from models from `sklearn` and `xgboost`. 
+`shaprpy` knows how to explain predictions from models from `sklearn`, `keras` and `xgboost`. 
 For other models, one can provide a custom `predict_model` function (and optionally a custom `get_model_specs`) to `shaprpy.explain`.
 
 See `/examples` for runnable examples, including an example of a custom PyTorch model.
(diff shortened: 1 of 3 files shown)


martinju:
Thanks, will look at this next week. Please try to see if it is possible to define a `get_model_specs ` function, with at least the feature names. This would make it much more secure to use in practice.

rawanmahdi:
I looked into implementing the `get_model_specs` function some more, and I asked on stack overflow if there was a way of obtaining feature names. The answer I got: 

"No, unlike sklearn there isn't a Keras/Tensorflow equivalent for obtaining feature names. This is because tensorflow-keras models focus more on the shape and dimensions of the input tensors rather than the individual features. The model utilizes the input features as tensors, regardless of name."

I think the only work around for validating between the dataset and the model would be to check `feature_size` or `feature_type`, however from what I can see that would require updating R internals as well. I'm not too familiar with R, however I'll look into it for a seperate PR.

martinju (approved):
I think this looks OK. Added a seed for the example to get reproducable results (although I am sure it also depends on python/keras/tensorflow versions. Not getting feature labels is something I guess we have to live with, I understand this is less common in python than R. The only issue is the somewhat confusing warning messages coming partly from python and partly from R saying that the model is not natively supported etc. Will have to find a way to bypass that eventually.

PR #349, opened 2023-06-27
Fixed bug with "keep_samp_for_vS"…

LHBO:
…in the internal function postprocess_vS_list() which occured if keep_samp_for_vS = TRUE in the explain() function. Previously, the code did not take into consideration that vS_list became a list of lists when keep_samp_for_vS = TRUE and therefore extracted incorrect names.


martinju (approved):
I added a test of the output when keep_samp_for_vS = TRUE to ensure a similar bug is not introduced in the future.

Ready to merge once tests are complete.

PR #353, opened 2023-08-07
Fixed a bug with the `gaussian` approach not using the user-provided …

LHBO:
…values for the `gaussian.mu` and `gaussian.cov_mat` arguments, and added a test function. Discuss with Martin what to do with the 2nd and 3rd function arguments, as they are unnecessary, and the roxygen2 documentation.

Previously, the `setup_approach.gaussian` function did not extract the `gaussian.mu` and `gaussian.cov_mat` arguments from the `internal` list but rather let them be two separate arguments in the function call. 
However, they never reached the function as they were part of the `...` arguments, which were not a part of the `setup_computation(internal, model, predict_model)` call in the `explain` function, which calls the `setup_approach.gaussian` function.

### DIFF
--- a/R/approach_gaussian.R
+++ b/R/approach_gaussian.R
@@ -14,11 +14,10 @@
 setup_approach.gaussian <- function(internal,
                                     gaussian.mu = NULL,
                                     gaussian.cov_mat = NULL, ...) {
+
   parameters <- internal$parameters
   x_train <- internal$data$x_train
   feature_specs <- internal$objects$feature_specs
-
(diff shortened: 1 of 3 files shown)


martinju (approved):
Thanks for spotting this, fixing it and adding a test!
I did some minor edits to fix consistency and lint complaints. 
While the function arguments `gaussian.mu` etc are not used directly they are used to define defaults (which are inherited by the documentation of `explain`), see also the other `setup_approach.*` functions. I think this is an OK way to do this. Let me know if you have a better idea. 

Feel free to merge.

LHBO (commented):
Reviewed and agree with your alterations.

PR #356, opened 2023-08-24
Bugfix in `prepare_data()` related to vector of approaches. When usin…

LHBO:
…g several approaches; the old version only used the first approach. I verified this by adding a temporary print in each `prepare_data.approach()` function and saw that only the first approach in `internal$parameters$approach` was printed. 
It is not possible to spot the logical error in any other way to the best of my knowledge, hence I have not been able to make a script demonstrating the error, and I have made no test.

We could, in all `prepare_data.approach()` functions, add an extra column in the returned data.table indicating the used approach to generate the MC samples. This would make it easier to test that the specified approaches were indeed used.

Remove code comments before the pull request is accepted, and maybe Martin has a better solution to get the approach(?).

Also updated roxygen2 for the function, as it seemed that it was reflecting the old version of shapr(?) due to arguments that are no longer present.

However, we then get a warning when creating the roxygen2 documentation. Discuss some solutions as comments below. Discuss with Martin.


LHBO:
It seems that the checks fail as the tests say that the combined approach now gives different results/explanations compared to what it did before. This might be a good thing because it illustrates that the updated version uses different approaches.

martinju:
Well done spotting this! Sorry you had to deal with this. As I guess you realized, I had most of this setup up, but apparently I had completely forgotten to handle it in prepare_data. 

I verified manually that your fix works as intended. As you wrote, I think it is hard to make a test for this (the mistake was to make an output test with the wrong output in the first place). An error was introduced in the output_lm_mixed_comb test, so that needs to be handled. The others test failures are purely based on the incorrect output test being updated (I will update those test files once the output_lm_mixed_comb error is fixed). Feel free to take a look at what is going on with output_lm_mixed_comb, otherwise I will do it later this week.

We inherit documentation from one/two sources when using `internal` to avoid copying text in lots of places (which makes maintenance more cumbersome). Thus, you should not write documentation for internal in other functions, so we should remove those again. The warning is that you say you want to inherit function documentation, but all input arguments are already documented.

LHBO:
The error in output_lm_mixed_comb is due to the same thing as in #362. Have you seen/read it?

R/approach.R
```diff
@@ -80,12 +80,8 @@ prepare_data <- function(internal, index_features = NULL, ...) {
 
   # Check if the user provided one or several approaches.
   if (length(approach) > 1) {
-    # Figure out which approach we should use for the provided `index_features` (i.e., coalition indices).
```
LHBO: Agree with better/less detailed comment. 
Gave a detailed comment to make it easier for you to understand the logic.

LHBO:
Have solved the bug at its origin. Will update soon.

LHBO:
Now I did something I did not intend to do...

LHBO:
The check fails, but I think that is due to old test-values? 
I'll wait and see what Martin says :)

R/approach_independence.R
```diff
@@ -44,22 +60,30 @@ prepare_data.independence <- function(internal, index_features = NULL, ...) {
...
   x_explain0_mat <- as.matrix(x_explain0)
 
+  # Get coalition indices.
+  # We repeat each coalition index `min(n_samples, n_train)` times. We use `min`
+  # as we cannot sample `n_samples` unique indices if `n_train` is less than `n_samples`.
   index_s <- rep(seq_len(nrow(S0)), each = min(n_samples, n_train))
-  w <- 1 / n_samples # Yes, not n_samples0
-
-  n_col <- n_explain
+  w0 <- 1 / min(n_samples, n_train) # The inverse of the number of samples being used in practice
```
martinju, line 71: I went for this for consistency. It does actually make a tiny difference what we use here, as the edge cases (full and empty set) are weighted 1, so their relative weight compared to those will differ if we choose something else.

R/approach_independence.R
```diff
@@ -71,24 +95,32 @@ prepare_data.independence <- function(internal, index_features = NULL, ...) {
...
+      # We extract the categories/levels for `this` categorical features
       this_levels <- level_list[[this]]
-      dt[, (this) := as.factor(get(this))]
-      levels(dt[[this]]) <- this_levels
+
+      # Then we convert `this` from integers to a factor where we specify that the `levels`
+      # go from 1 to the number of levels/categories (i.e., the max integer). We need to
+      # specify this in case not all the levels are presents in `dt` which can happen for rare
+      # levels. We set the `labels` for each of the `levels` to be the original categories
+      dt[, (this) := factor(get(this), levels = seq_along(this_levels), labels = this_levels)]
```
martinju, line 120: Looks good, thanks!

R/documentation.R
```diff
@@ -31,6 +31,10 @@ default_doc <- function() {
 #' Exported documentation helper function.
 #'
 #' @param internal Not used.
+#'
+#' @param index_features Positive integer vector. Specifies the indices of combinations to
+#' apply to the present method. `NULL` means all combinations. Only used internally.
+#'
```
martinju, line 37: Note to Lars: Added this to default doc instead.

R/approach.R
```diff
@@ -43,31 +39,13 @@ setup_approach.combined <- function(internal, ...) {
...
-#' `param ... Currently not used.` below, or remove rdname from the subfunctions and rather
-#  use inherateparams.
-#'
-#'
-#' @param internal List. Holds all parameters, data, functions and computed objects used
-#' within \code{\link[=explain]{explain()}}. The list contains one or more of the elements
-#' `parameters`, `data`, `objects`, `output`.
-#' @param index_features Positive integer vector. Specifies the indices of combinations to
-#' apply to the present method. `NULL` means all combinations. Only used internally.
-#'
```
martinju: As mentioned earlier, we use inheritParams to avoid copying. I inherited from default_doc_explain (rather than default_doc) as the documentation of explain is affected by this doc, and for that the explain function, it can be confusing if one is told to pass the internal object (hence "not used" although it is not strictly correct for the prepare_data function).

R/approach.R
```diff
@@ -43,31 +39,13 @@ setup_approach.combined <- function(internal, ...) {
 
 #' Generate data used for predictions and Monte Carlo integration
 #'
-#' @note
```
martinju: Also, in the future: put stuff like this in a comment here on github instead.

martinju (approved):
I think this good now and ready to merge, but please take a quick look at my latest edits.

R/approach_independence.R
LHBO: That makes sense, and I agree with you there. I like the way you made it in the final version.

R/approach_independence.R
LHBO: You are welcome :)

R/documentation.R
LHBO: Good, thank you. :) (Previously, I did not want to change that file without your consent, hence I tried to keep the changes locally, but now I know that I should have put it here)

LHBO (commented):
I have gone through the changes and agree with the alterations you have made to my proposed fixes.

R/approach.R
LHBO: Noted. Makes more sense to put it here than in the code...

R/approach.R
LHBO: Thank you for clarifying this, and it makes sense. I will try to remember this going forward.

PR #357, opened 2023-08-24
Lars/add MSEv eval criterion

LHBO:
Implemented the Mean Squared Error (MSEv evaluation criterion) of the contribution function v(s) as proposed by [Frye et al. (2019)](https://arxiv.org/pdf/2006.01272.pdf) and used by [Olsen et al. (2022)](https://www.jmlr.org/papers/volume23/21-1413/21-1413.pdf).

This evaluation criterion is computed for all approaches as long as `internal$parameters$output_size == 1`.

Have also added plot functions and included a section in the vignette.

### DIFF
--- a/R/finalize_explanation.R
+++ b/R/finalize_explanation.R
@@ -24,12 +24,10 @@ finalize_explanation <- function(vS_list, internal) {
 
   # internal$timing$shapley_computation <- Sys.time()
 
-
-  # Clearnig out the tmp list with model and predict_model (only added for AICc-types of empirical approach)
+  # Clearing out the tmp list with model and predict_model (only added for AICc-types of empirical approach)
   internal$tmp <- NULL
 
   internal$output <- processed_vS_list
-
 
   output <- list(
     shapley_values = dt_shapley,
@@ -37,6 +35,19 @@ finalize_explanation <- function(vS_list, internal) {
     pred_explain = p
   )
   attr(output, "class") <- c("shapr", "list")
+
+  # Compute the MSEv evaluation criterion if the output of the predictive model is a scalar.
+  # TODO: check if it makes sense for output_size > 1.
+  if (internal$parameters$output_size == 1) {
+    # Compute the MSEv evaluation criterion
+    MSEv_evaluation_criterion = compute_MSEv_evaluation_criterion(
+      internal = internal,
+      processed_vS_list = processed_vS_list,
+      p = p,
+      exclude_empty_and_grand_coalition = FALSE,
+      return_as_dt = TRUE)
+    output$MSEv_evaluation_criterion = MSEv_evaluation_criterion
+  }
 
   return(output)
 }
@@ -103,7 +114,7 @@ get_p <- function(dt_vS, internal) {
 #' Compute shapley values
 #' @param explainer An `explain` object.
 #' @param dt_vS The contribution matrix.
-#' @return A `data.table` with shapley values for each test observation.
+#' @return A `data.table` with Shapley values for each test observation.
 #' @export
 #' @keywords internal
 compute_shapley_new <- function(internal, dt_vS) {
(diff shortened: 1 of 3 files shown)


LHBO:
The checks fail, and it seems like it is because the vignette does not find the new `make_MSEv_evaluation_cirterion_plots()` function. I do not understand why that happens. The function is exported. Do I need to add `shapr::` before the function call? But that is not needed when er call the `explain()` function.

LHBO:
The errors now seem to be related to some test failures. These tests compare with the previous output from the `explain()` function, but now the output from the `explain()` function also includes the MSEv evaluation criterion. Note 100% sure. Discuss with Martin if I return the MSEv correct place. See the vignette for usage of the MSEv.

LHBO:
I believe that this branch is done.  However, there are something to note/discuss.

1. Lintr complains about long variable names, but I personally like them. It also complains about the use of curly brackets, but they were added by styler.
2. I had to merge the other pull requests #356 and #359 into this because of the bugs in `shapr`. Otherwise, I would have created test functions with incorrect truth.
3. Thus if Martin changes the PR above, then that should also be incorporated in here.

R/plot.R
```diff
@@ -780,3 +780,885 @@ make_waterfall_plot <- function(dt_plot,
...
+#'
+#' @param explanation_list A list of [shapr::explain()] objects applied to the same data and model.
+#' If the entries in the list is named, then the function use these names. Otherwise, it defaults to
+#' the approach names (with integer suffix for duplicates) for the explanation objects in `explanation_list`.
+#' @param index_explicands Integer vector. Which of the explicands (test observations) to plot.
+#' E.g. if you have explained 10 observations using [shapr::explain()], you can generate a plot for the
+#' first 5 observations/explicands and the 10th by setting `index_x_explain = c(1:5, 10)`.
+#' @param index_combinations Integer vector. Which of the coalitions (combinations) to plot.
+#' E.g. if you used `n_combinations = 16` in [explain()], you can generate a plot for the
+#' first 5 coalitions and the 10th by setting `index_x_explain = c(1:5, 10)`.
```
LHBO, line 801: Typo

martinju:
> I believe that this branch is done. However, there are something to note/discuss.
> 
>     1. Lintr complains about long variable names, but I personally like them. It also complains about the use of curly brackets, but they were added by styler.
> 
>     2. I had to merge the other pull requests [Bugfix in `prepare_data()` related to vector of approaches. When usin… #356](https://github.com/NorskRegnesentral/shapr/pull/356) and [Harmonize batch distribution ++ #359](https://github.com/NorskRegnesentral/shapr/pull/359) into this because of the bugs in `shapr`. Otherwise, I would have created test functions with incorrect truth.
> 
>     3. Thus, if Martin changes the PR above, then that should also be incorporated in here.
> 
>     4. The vignette is not able to be created... It runs locally on my computer (except for the `future` section).

I updated the branch and fixed failing tests and vignette. For your comments:

1. If you do ..._evaluation_criterion_... -> ..._eval_... or ..._eval_crit_... I think most of these will go away. Using such long names makes the code hard to read, so I think we should obey these.
2. Good, this should be OK now, I think.
3. Yep.
4. Runs locally for me too (now). Lets see what GHA says.

R/finalize_explanation.R
```diff
@@ -24,20 +24,35 @@ finalize_explanation <- function(vS_list, internal) {
...
+    output$MSEv_eval_crit <- compute_MSEv_eval_crit(
+      internal = internal,
+      processed_vS_list = processed_vS_list,
+      p = p,
+      exclude_empty_grand_coalitions = ifelse(!is.null(internal$parameters$exclude_empty_grand_coalitions),
+                                              internal$parameters$exclude_empty_grand_coalitions,
+                                              FALSE),
+      use_shapley_weights = ifelse(!is.null(internal$parameters$use_shapley_weights),
+                                   internal$parameters$use_shapley_weights,
+                                   FALSE),
```
martinju, line 51: - parameters in internal should be declared in the top of the function for clarity and readability on what is used 
	within the function. Just write e.g. aa = internal$parameters$blabla$aa on the top and use just aa later. 
- consider changing name on exclude_empty_grand_coalitions and use_shapley weights
- where are these two parameters defined? are they added to the output of explain()? I dont think so

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
 
   return(dt_kshap)
 }
+
+
+
+#' Mean Squared Error of the Contribution Function `v(S)`
+#'
+#' @param internal List with the different parameters, data and functions used internally in the [explain()] function.
```
martinju, line 176: Do not rewrite documentation of common arguments, use #' @inheritParams default_doc instead.

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
...
+#' @param processed_vS_list List of the processed contribution function estimates.
+#' Output from the \code{shapr:::postprocess_vS_list()} function.
+#' @param p Numeric vector with the predicted responses for the observations which are to be explained.
+#' Output from the \code{shapr:::get_p()} function.
+#' @param exclude_empty_grand_coalitions Boolean. If `TRUE`, we exclude the empty and grand coalitions
+#' when computing the MSEv evaluation criterion. This is reasonable as they are identical for all methods, i.e.,
+#' their contribution function is independent of the used method as they are special cases not effected by
+#' the used method. If `TRUE`, we exclude the empty and grand coalitions.
```
martinju, line 184: The last sentence should rather be about FALSE.

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
...
+#' @param return_as_dt Boolean. If the computed MSEv evaluation criterion is
+#' to be returned as \code{\link[data.table]{data.table}}.
+#' @param use_shapley_weights Boolean. If `TRUE`, then we use the Shapley kernel weights to weight the
+#' different coalitions before taking the mean. If `FALSE`, we use regular mean with uniform weights.
```
martinju, line 188: write (default) in parenthesis for the default. this also goes for the above

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
...
+#'   processed_vS_list = explanation$internal$output,
+#'   exclude_empty_grand_coalitions = FALSE,
+#'   return_as_dt = FALSE
+#' )
+#'
+#' @author Lars Henry Berge Olsen
+#' @export
+compute_MSEv_eval_crit <- function(internal,
+                                   processed_vS_list,
+                                   p = shapr:::get_p(processed_vS_list$dt_vS, internal),
```
martinju, line 293: Avoid using `shapr:::`
We should also think through how to do this when using the function outside of shapr, as one should not be required to call an internal function

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
 
   return(dt_kshap)
 }
+
+
+
+#' Mean Squared Error of the Contribution Function `v(S)`
```
martinju, line 174: General comment: Try to shorten the code a bit. Comments are good for getting an overview of what is done, but we don't need it for every single line (normalizing + extracting internal parameters are examples of obvious lines).

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
...
+  if (use_shapley_weights) {
+    if (!exclude_empty_grand_coalitions) {
+      warning(paste0("We do not reccomend seting `exclude_empty_grand_coalitions = FALSE` when ",
+                     "`use_shapley_weights = TRUE`, as the weights of the empty and full coalitions ",
+                     "drastically outweights the other coalitions. This causes the MSEv criterion to be ",
+                     "more or less identical for all approaches."))
+    }
+
+    # We are using the weighted
+    averaging_weights = internal$objects$X$shapley_weight
```
martinju, line 326: I guess you can simplify the code by defining averaring_weights conditional on use_shapley_weights in one line, to then reuse the code in lines 340-345 for both case.

R/finalize_explanation.R
```diff
@@ -153,3 +168,238 @@ compute_shapley_new <- function(internal, dt_vS) {
...
+    # Compute the overall mean squared error averaged over both the coalitions and explicands.
+    MSEv_eval_crit <- mean(dt_squared_diff)
+  }
+
+  # Set the names entires in the array
+  names(MSEv_eval_crit_explicand) <- paste0("id_", seq(n_explain))
+  names(MSEv_eval_crit_combination) <- paste0("id_combination_", id_combination_numbers)
+  names(MSEv_eval_crit_combination_sd) <- paste0("id_combination_", id_combination_numbers)
+
+  # If we are to return the results as data.table, then we overwrite the previous results.
```
martinju, line 370: Not use what you mean by this, or if these is an issue by always returning the results as a data.table (or always a list). Do we need 2 options here?

martinju (changes requested):
Comments related to the compute_MSEv_eval_crit function. Will look at the plotting functions later. BTW, almost none of the comments are crucial.

R/finalize_explanation.R
LHBO: Added default to the documentation.

R/finalize_explanation.R
LHBO: - In the new version I have declared the parameters at the begining.
- I have changed the names `exclude_empty_grand_coalitions` and `use_shapley weights` to `MSEv_skip_empty_full_comb` and `MSEv_uniform_comb_weights`, respectively. 
- They were poorly defined in the sense that the user could send them as `...` parameters in `explain()`. If they did not, then we used it as specified above. They were added to the output of `explain()`, IF the user specified them. I have now made them as arguments to `explain()`, made a proper test for valid input in `setup()` and added documentation to `explain()`.

R/finalize_explanation.R
LHBO: I had forgotten about that one. Thanks!

R/finalize_explanation.R
LHBO: Correct. I have updated the documentation + used `#' @inheritParams default_doc` and `#' @inheritParams explain`

R/finalize_explanation.R
LHBO: I have removed the `shapr:::p` and examples and rather done the same as in `compute_shapley_new()`.

R/finalize_explanation.R
LHBO: They were a list first, but then I wanted to add some extra information. E.g `features` for `MSEv_eval_crit_combination` in case we are using sampling of coalitions since it is then nice to see which coalitions `id_combination` number `x` belongs to. I have removed the option and always return a list of `data.tables`.

R/plot.R
```diff
@@ -780,3 +780,519 @@ make_waterfall_plot <- function(dt_plot,
...
+#'   prediction_zero = prediction_zero,
+#'   n_samples = 1e2
+#' )
+#'
+#' # Combined approach
+#' explanation_combined <- explain(
+#'   model = model,
+#'   x_explain = x_explain,
+#'   x_train = x_train,
+#'   approach = c("gaussian", "independence", "empirical"),
```
martinju, line 905: swap empirical with ctree

R/plot.R
```diff
@@ -780,3 +780,519 @@ make_waterfall_plot <- function(dt_plot,
...
+#'
+#' # Create a list of explanations without names
+#' explanation_list_unnamed <- list(
+#'   explanation_independence,
+#'   explanation_empirical,
+#'   explanation_gaussian_1e1,
+#'   explanation_gaussian_1e2,
+#'   explanation_ctree,
+#'   explanation_combined
+#' )
```
martinju, line 918: skip

R/plot.R
```diff
@@ -780,3 +780,519 @@ make_waterfall_plot <- function(dt_plot,
...
+#'   # We can specify which test observations or combinations to plot
+#'   make_MSEv_eval_crit_plots(explanation_list_named,
+#'                             make_MSEv_comb_and_explicand = TRUE,
+#'                             index_x_explain = c(1, 3:4, 6),
+#'                             N_sd_error_bars = 1)$MSEv_explicand_bar
+#'   make_MSEv_eval_crit_plots(explanation_list_named,
+#'                             make_MSEv_comb_and_explicand = TRUE,
+#'                             id_combination = c(3, 4, 9, 13:15),
+#'                             N_sd_error_bars = 1)$MSEv_combination_bar
+#' }
```
martinju, line 1013: I think this is too much and make the help files very long. And even if the plotting is done outside of shapr it takes a bit of time. I think we should reduce it significantly. We can instead put this in a script or something and put it on github.

R/plot.R
```diff
@@ -780,3 +780,519 @@ make_waterfall_plot <- function(dt_plot,
...
+#' @inheritParams default_doc
+#'
+#' @param explanation_list A list of [shapr::explain()] objects applied to the same data and model.
+#' If the entries in the list are named, then the function use these names. Otherwise, they default to
+#' the approach names (with integer suffix for duplicates) for the explanation objects in `explanation_list`.
+#' @param id_combination Integer vector. Which of the combinations (coalitions) to plot.
+#' E.g. if you used `n_combinations = 16` in [explain()], you can generate a plot for the
+#' first 5 combinations and the 10th by setting `index_x_explain = c(1:5, 10)`.
+#' @param geom_col_width Numeric. Bar width. By default, set to 90% of the [ggplot2::resolution()] of the data.
+#' @param N_sd_error_bars Positive numeric (default is `1`). The number of standard deviation to scale error bars in
```
martinju, line 804: Mention that 0 removes it.

R/plot.R
```diff
@@ -780,3 +780,519 @@ make_waterfall_plot <- function(dt_plot,
...
+                                          index_x_explain = index_x_explain,
+                                          id_combination = id_combination)
+  MSEv_dt = MSEv_dt_list$MSEv
+  MSEv_explicand_dt = MSEv_dt_list$MSEv_explicand
+  MSEv_combination_dt = MSEv_dt_list$MSEv_combination
+
+  # Warnings related to the error bars
+  if (!is.null(N_sd_error_bars) && N_sd_error_bars > 0) {
+    if (n_explain < 30) {
+      message(sprintf("The error bars might be wide as the number of observations to explain is only %d.\n",
```
martinju, line 1053: not sure I follow. So what you are plotting is sd/sqrt(n)? clarify in documentation

martinju (commented):
see some comments

R/plot.R
LHBO: Specified that setting `levels = NULL` in the new version removes the CI.

R/plot.R
LHBO: Shorted the examples.  Put the other examples into the `example_plot_MSEv.R` file inside the `shapr/inst/scripts` folder.

R/plot.R
LHBO: Removed the unnamed list and the corresponding example

R/plot.R
LHBO: Swapped emp with ctree.

R/explain.R
```diff
@@ -82,8 +82,21 @@
 #' disabled for unsupported model classes.
 #' Can also be used to override the default function for natively supported model classes.
 #'
+#' @param MSEv_skip_empty_full_comb Logical. If `TRUE`, we exclude the empty and grand
+#' combinations/coalitions when computing the MSEv evaluation criterion. This is reasonable as they are identical
+#' for all methods, i.e., their contribution function is independent of the used method as they are special cases not
+#' effected by the used method. If `FALSE` (default), we include the empty and grand combinations/coalitions.
```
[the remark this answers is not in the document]
LHBO: Removed the parameter and updated the documentation and tests.

R/plot.R
LHBO: I have added clarifying text in the documentation of the MSEv plot function.

LHBO: I have also added some sentences to the vignette

LHBO:
> see some comments

Have rewritten the code based on your feedback. 
I have run `lintr::lint_dir()`, `styler:::style_dir()`, and `devtools::run_examples()`. 
We I run `testthat()`, then I get a lot of errors which are due to the objects being changed due to e.g. new names.

martinju:
I think this looks good now. I have some minor suggestions for name changes:

make_MSEv_eval_crit_plots -> plot_MSEv_eval_crit
And in this function: 
level = 0.95 -> CI_level = ifelse(length(explanation_list[[1]]$pred_explain)<20, NULL, 0.95) [name change + don't plot CI if there are fewer than (say) 20 obs]
make_MSEv_comb_and_explicand -> plot_type = "overall" [a character vector specifying which plots to produce, can take "overall", "comb" and "explicand"]

What do you think, @LHBO ?

martinju (approved):
I fixed a few minor issues. I think this is good for now. 
Feel free to merge if you agree.

PR #358, opened 2023-08-24
Fixed bug where the order of the columns where different and resultin…

LHBO:
…g in rbinding incorrect columns together. Only occurred when a batch only contained the grand coalition.

Both fixed the bug at its source and added `use.names=TRUE` in `rbindlist()` for extra safety.

### DIFF
--- a/R/compute_vS.R
+++ b/R/compute_vS.R
@@ -120,7 +120,13 @@ batch_prepare_vS <- function(S, internal) {
     } else {
       dt <- NULL # Special case for when the batch only include the largest id
     }
-    dt_max <- data.table(x_explain, id_combination = max_id_combination, w = 1, id = seq_len(n_explain))
+    # Remove comment in the end.
+    # Lars: Fixed incorrect order that only made an effect for a batch containing only the max_id_combination,
+    # as then dt is NULL, so when we rbind dt with dt_max we kept the order of columns from dt_max.
+    # Otherwise, when dt is NOT NULL, then the order would be the same as in dt.
+    # So beforehand we ended up rbindin id_combination and a feature, and all features where shifted
+    # one column to the left. Date 15-09-23.
+    dt_max <- data.table(id_combination = max_id_combination, x_explain, w = 1, id = seq_len(n_explain))
     dt <- rbind(dt, dt_max)
     setkey(dt, id, id_combination)
   }
--- a/R/compute_vS.R
+++ b/R/compute_vS.R
@@ -120,12 +120,6 @@ batch_prepare_vS <- function(S, internal) {
     } else {
       dt <- NULL # Special case for when the batch only include the largest id
     }
-    # Remove comment in the end.
-    # Lars: Fixed incorrect order that only made an effect for a batch containing only the max_id_combination,
-    # as then dt is NULL, so when we rbind dt with dt_max we kept the order of columns from dt_max.
-    # Otherwise, when dt is NOT NULL, then the order would be the same as in dt.
-    # So beforehand we ended up rbindin id_combination and a feature, and all features where shifted
-    # one column to the left. Date 15-09-23.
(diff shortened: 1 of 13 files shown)


R/compute_vS.R
```diff
@@ -120,7 +120,13 @@ batch_prepare_vS <- function(S, internal) {
...
       dt <- NULL # Special case for when the batch only include the largest id
     }
-    dt_max <- data.table(x_explain, id_combination = max_id_combination, w = 1, id = seq_len(n_explain))
+    # Remove comment in the end.
+    # Lars: Fixed incorrect order that only made an effect for a batch containing only the max_id_combination,
+    # as then dt is NULL, so when we rbind dt with dt_max we kept the order of columns from dt_max.
+    # Otherwise, when dt is NOT NULL, then the order would be the same as in dt.
+    # So beforehand we ended up rbindin id_combination and a feature, and all features where shifted
+    # one column to the left. Date 15-09-23.
+    dt_max <- data.table(id_combination = max_id_combination, x_explain, w = 1, id = seq_len(n_explain))
```
martinju, line 129: Please add a test with an example which gives this error. If you don't find a good place for it, feel free to create a new file test-various-bugfixes or so and throw it in there.

R/finalize_explanation.R
```diff
@@ -66,7 +66,22 @@ postprocess_vS_list <- function(vS_list, internal) {
...
+    # One could potentially also solve this bug by determine why the column order is different in
+    # the first place. Date 15-09-23.
+    # Message 2:
+    # The origin of the bug was in `batch_prepare_vS` function and I have now fixed it, so the
+    # column order is of the correct order. Thus, could remove use.names=TRUE, but it might
+    # be good to have it there. The rbind.data.table function has sets it to TRUE by default,
+    # as described in news for v1.12.2.
+    # https://cran.r-project.org/web/packages/data.table/news/news.html
+    # Date 15-09-23.
+    dt_samp_for_vS <- rbindlist(lapply(vS_list, `[[`, 2), use.names = TRUE)
```
martinju, line 83: I agree, we should keep use.names = TRUE even though you have fixed the initial bug.

R/compute_vS.R
LHBO: Will add it tomorrow in a proper fashion:

`library(xgboost)
library(data.table)

data("airquality")
data <- data.table::as.data.table(airquality)
data <- data[complete.cases(data), ]

x_var <- c("Solar.R", "Wind", "Temp", "Month")
y_var <- "Ozone"

ind_x_explain <- 1:6
x_train <- data[-ind_x_explain, ..x_var]
y_train <- data[-ind_x_explain, get(y_var)]
x_explain <- data[ind_x_explain, ..x_var]

model <- xgboost::xgboost(
  data = as.matrix(x_train),
  label = y_train,
  nround = 20,
  verbose = FALSE
)

# Specifying the phi_0, i.e. the expected prediction without any features
p0 <- mean(y_train)

# Computing the actual Shapley values with kernelSHAP accounting for feature dependence using
# the gaussian approach
# Note that we get an warning due to the bug.
explanation <- explain(
  model = model,
  x_explain = x_explain,
  x_train = x_train,
  approach = "gaussian",
  prediction_zero = p0,
  n_samples = 2,
  n_batches = 15,
  keep_samp_for_vS = TRUE
)

# Here we can see that id combination suddenly is a not an integer. 
# This is fixed after the bugfix has beed applied
explanation$internal$output$dt_samp_for_vS
`

LHBO:
Looked at the new test, and it looks good. (testing out if this I was supposed to comment or if I was supposed to comment inside the commit)

LHBO (commented):
Looked at the new test, and it looks good. (testing out commenting from inside the commit)

martinju (approved):

martinju:
> Looked at the new test, and it looks good. (testing out commenting from inside the commit)

Don't think it really matters.

PR #359, opened 2023-08-24
Fixed bug in get_default_n_batches function where shapr sets the numb…

LHBO:
…er of batches (when not provided in the explain function call) to a number which then throws an error in the check_n_batches function. Logical error as get_default_n_batches could previously set n_batches to a larger value than n_combinations - 1. Subtract one as the check_n_batches function specifies that n_batches must be strictly less than n_combinations.

The bug occurred for example for:
```
# The other stuff here is the variables from the first code block in the vignette.
> explanation <- explain(
+   model = model,
+   x_explain = x_explain,
+   x_train = x_train,
+   approach = "gaussian",
+   prediction_zero = p0,
+   n_combinations = 8
+ )
Note: Feature classes extracted from the model contains NA.
Assuming feature classes from the data are correct.

Setting parameter 'n_batches' to 10 as a fair trade-off between memory consumption and computation time.
Reducing 'n_batches' typically reduces the computation time at the cost of increased memory consumption.

Error in check_n_batches(internal) : 
  `n_batches` (10) must be smaller than the number of feature combinations/`n_combinations` (8)
```

### DIFF
--- a/R/setup.R
+++ b/R/setup.R
@@ -693,15 +693,13 @@ get_default_n_batches <- function(approach, n_combinations) {
     suggestion <- ceiling(n_combinations / 10)
     this_min <- 10
     this_max <- 1000
-    min_checked <- max(c(this_min, suggestion))
-    ret <- min(c(this_max, min_checked))
   } else {
     suggestion <- ceiling(n_combinations / 100)
     this_min <- 2
     this_max <- 100
-    min_checked <- max(c(this_min, suggestion))
-    ret <- min(c(this_max, min_checked))
   }
+  min_checked <- max(c(this_min, suggestion))
+  ret <- min(c(this_max, min_checked, n_combinations - 1))
   message(
     paste0(
       "Setting parameter 'n_batches' to ", ret, " as a fair trade-off between memory consumption and ",
--- a/R/setup.R
+++ b/R/setup.R
@@ -217,7 +217,7 @@ check_n_batches <- function(internal) {
 
   if (n_batches >= actual_n_combinations) {
     stop(paste0(
-      "`n_batches` (", n_batches, ") must be smaller than the number feature combinations/`n_combinations` (",
+      "`n_batches` (", n_batches, ") must be smaller than the number of feature combinations/`n_combinations` (",
       actual_n_combinations, ")"
     ))
   }


LHBO:
The tests failed as I changed the printout. I added the word **of** as I thought that was more grammatically correct. But one can just remove it again. The intention of the sentence is clear.

The new printout is:
`n_batches` (32) must be smaller than the number **of** feature combinations/`n_combinations` (32)

R/setup.R
```diff
@@ -217,7 +217,7 @@ check_n_batches <- function(internal) {
 
   if (n_batches >= actual_n_combinations) {
     stop(paste0(
-      "`n_batches` (", n_batches, ") must be smaller than the number feature combinations/`n_combinations` (",
+      "`n_batches` (", n_batches, ") must be smaller than the number of feature combinations/`n_combinations` (",
```
LHBO, line 220: This makes the previous tests fail, as now the error messages are different. Let Martin decide if we keep this alteration or not.

LHBO:
Should not be merged yet, as another bug occurred when I was working on combined approaches.

```
explain_numeric = explain(
  model = model_lm_numeric,
  x_explain = x_explain_numeric,
  x_train = x_train_numeric,
  approach = c("independence", "empirical", "gaussian", "copula", "empirical"),
  prediction_zero = p0,
  n_batches = NULL,
  timing = FALSE)
Setting parameter 'n_batches' to 2 as a fair trade-off between memory consumption and computation time.
Reducing 'n_batches' typically reduces the computation time at the cost of increased memory consumption.


explain_numeric$internal$objects$S_batch
$`1`
[1]  2  3  4  5  6 32

$`2`
 [1]  7  8  9 10 11 12 13 14 15 16

$`3`
 [1] 17 18 19 20 21 22 23 24 25 26

$`4`
[1] 27 28 29 30 31

> explain_numeric$internal$parameters$n_batches
[1] 2

```

LHBO:
There are also some other additional bugs with the combined approaches, e.g., that setting seed does not work.

LHBO:
Looks like some test fails. Will look at that tomorrow.

martinju:
I looked at the tests here and almost all of them are OK (just adding two new objects), except for the test involving the independence approach, where 'something' is off slightly, causing a small numerical change (10th decimal place or something). It is probably fine, but would be good to know what causes it. That will be easier to see once #356 is merged as this PR contains all those changes as well, making it harder to spot what is actually changed here.

LHBO:
> I looked at the tests here and almost all of them are OK (just adding two new objects), except for the test involving the independence approach, where 'something' is off slightly, causing a small numerical change (10th decimal place or something). It is probably fine, but would be good to know what causes it. That will be easier to see once #356 is merged as this PR contains all those changes as well, making it harder to spot what is actually changed here.

I have merged #356 into this PR now. Will try to look more into it.

LHBO:
La til noen comments

LHBO (commented):
added comments review

R/setup_computation.R
```diff
@@ -632,6 +633,57 @@ create_S_batch_new <- function(internal, seed = NULL) {
...
+      # Ensure that the number of batches is not lower than `n_batches`.
+      # Add one batch to the approach with most coalitions per batch
+      while (sum(batch_count_dt$n_batches_per_approach) < n_batches) {
+        approach_to_add_batch <- which.max(batch_count_dt$n_S_per_approach /
+                                            batch_count_dt$n_batches_per_approach)
+        batch_count_dt$n_batches_per_approach[approach_to_add_batch] <-
+          batch_count_dt$n_batches_per_approach[approach_to_add_batch] + 1
+      }
+    }
+
```
martinju, line 686: @LHBO Not sure whether this extra complexity is really worth it to cover the edge cases where the n_batches does not add up. An alternative is to provide a warning that the number of batches is changed (if specified byt he user), or otherwise just change it. What do you think?

martinju:
The failing tests on R4.2 and R4.1 (oldrel-1 and oldrel-2) seems to be due to a change on how errors are reported to the terminal. In the previous versions a missing input argument would not throw an error before one tried to access the missing argument, while in R4.3 an error is thrown immediately. This causes an error now due to an update in testthat (remember to update to v3.2.0 if you run locally) which now shows which function is throwing the error, which is then different in the two versions. I will try to look into how to get around this. One possibility is to simply ignore it as we know the reason for the error, it does not cause practical problems.

R/setup_computation.R
LHBO: @martinju, I agree that it is not very elegant code, but it does not take long to run and is only run once.
The two main reasons I added this is:

1. The first reason was to ensure that the number of batches adds up to `n_batches`.
2. The second reason, and likely the most important, is that if the user has specified `future::plan(multisession, workers = n_batches)`, then it would be nice that `shapr`actually used `n_batches`.

If this is not that important, then I am okay with just giving a warning.

R/setup.R
LHBO, line 1: We do not need the if-else anymore, so a simpler version of the code could be:

``` 
# Get the number of approaches, which is always either one or 
# one less than the number of features if a combination of approaches is used.
internal$parameters$n_approaches <- length(internal$parameters$approach)

# Get the number of unique approaches, as the same
# approach can be used for several feature combination sizes.
internal$parameters$n_unique_approaches <- length(unique(internal$parameters$approach))
```

LHBO (commented):
Propose some code simplification in `get_extra_parameters()`.

martinju (approved):
I think we are ready to merge this now (provided the checks pass)

PR #360, opened 2023-08-24
Fixed bug. All messages says "n_combinations is larger than or equal …

LHBO:
…to 2^m", but all the test only tested for "larger than". I.e., if the user specified n_combinations = 2^m in the call to shapr::explain, the function would not treat it as exact.

In the example below (the other stuff are the models and data from the vignette), we see that Shapley weights are sampled and that we are not using the exact version even though we use all of the coalitions.
```
> explanation <- explain(
+     model = model,
+     x_explain = x_explain,
+     x_train = x_train,
+     approach = "gaussian",
+     prediction_zero = p0,
+     n_combinations = 16
+ )
Note: Feature classes extracted from the model contains NA.
Assuming feature classes from the data are correct.

Setting parameter 'n_batches' to 10 as a fair trade-off between memory consumption and computation time.
Reducing 'n_batches' typically reduces the computation time at the cost of increased memory consumption.

> explanation$internal$objects$X$shapley_weight
 [1] 1000000       4       4       2       3       3       5       1       3       1       1       7       4       4       6 1000000
```

### DIFF
--- a/tests/testthat/test-setup.R
+++ b/tests/testthat/test-setup.R
@@ -1351,6 +1351,7 @@ test_that("Shapr with `n_combinations` >= 2^m uses exact Shapley kernel weights"
 
   # Create three explainer object: one with exact mode, one with
   # `n_combinations` = 2^m, and one with `n_combinations` > 2^m
+  # No message as n_combination = NULL sets exact mode
   expect_no_message(
     object = {
       explanation_exact <- explain(
@@ -1362,7 +1363,8 @@ test_that("Shapr with `n_combinations` >= 2^m uses exact Shapley kernel weights"
         n_samples = 2, # Low value for fast computations
         n_batches = 1, # Not related to the bug
         seed = 123,
-        n_combinations = NULL
+        n_combinations = NULL,
+        timing = FALSE
       )
     }
   )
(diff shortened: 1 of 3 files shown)


LHBO:
The check failed due to the cyclomatic complexity, it seems like.

LHBO:
Now several of the checks get cancelled. I am not sure why.

martinju (approved):
Good catch! I changed the tests a bit, checking for equality of the entire objects instead, and explicitly for the exact = TRUE and n_combinations = 32. Also fixed the linting. If you agree, feel free to merge once tests are done.

tests/testthat/test-setup.R
```diff
@@ -1345,6 +1345,93 @@ test_that("Error with too low `n_combinations`", {
...
+  expect_equal(
+    explanation_exact,
+    explanation_equal
+  )
+  expect_equal(
+    explanation_exact,
+    explanation_larger
+  )
+
+  # Explicitly check that exact mode is set and that n_combinations equals 2^ncol(x_explain_numeric) (32)
```
LHBO, line 1422: Smart to add this too!

LHBO (commented):
I agree with your changes

PR #361, opened 2023-08-24
Fixed visual bug in `plot.shapr()` which gave extra whitespace in the…

LHBO:
… strings "feature = value".

My proposed solution is to add trim whitespace, as we previously got extra white space for those features that have different numbers of digits. E.g., if a feature is "1000" for one observation and "5" for another, then we would previously get the strings "1000" and "   5", i.e., the latter/smaller one got three extra white spaces.


LHBO:
Lint complains about a lot of indentation of code I have not altered.

LHBO:
Also some of the test functions in `test-plot.R` now fails as the figures will differ as the extra whitespaces are removed.

LHBO:
I don't think I need to add extra test functions here as you have already written several. It is just that the "solution"/correct figure needs to be updated (?).

martinju (approved):

PR #364, opened 2023-09-04
Fix linting and stylr

martinju:
Disabled indentation linters as indentation don't agree with rstudio shortcuts for data.table and other edge cases. May enable it in the future. 
Also, too cumbersome to avoid the cyclocomp linting complaints so disabled that too.


LHBO (approved):
Reviewed the code and everything seems to be good.

martinju:
> 

We should delete to clean up things. If we want to back for some reason it can be restored (by clicking restore branch).

PR #366, opened 2023-12-04
Improving the efficiency of the Gaussian and (Gaussian) copula methods

### LINKED ISSUE #231: Improve efficency for Gaussian method
The implementation of the Gaussian method is far from optimal. 

This is a snippet from `prepare_data`: 
```
  for (i in seq(n_xtest)) {
    l <- lapply(
      X = features,
      FUN = sample_gaussian,
      n_samples = n_samples,
      mu = x$mu,
      cov_mat = x$cov_mat,
      m = ncol(x$x_test),
      x_test = x$x_test[i, , drop = FALSE]
    )
```
and this is `sample_gaussian`

```
sample_gaussian <- function(index_given, n_samples, mu, cov_mat, m, x_test) {

  # Check input
  stopifnot(is.matrix(x_test))

  # Handles the unconditional and full conditional separtely when predicting
  cnms <- colnames(x_test)
  if (length(index_given) %in% c(0, m)) return(data.table::as.data.table(x_test))

  dependent_ind <- (1:length(mu))[-index_given]
  x_test_gaussian <- x_test[index_given]
  tmp <- condMVNorm::condMVN(
    mean = mu,
    sigma = cov_mat,
    dependent.ind = dependent_ind,
    given.ind = index_given,
    X.given = x_test_gaussian
  )

  # Makes the conditional covariance matrix symmetric in the rare case where numerical instability made it unsymmetric
  if (!isSymmetric(tmp[["condVar"]])) {
    tmp[["condVar"]] <- Matrix::symmpart(tmp$condVar)
  }

  ret0 <- mvnfast::rmvn(n = n_samples, mu = tmp$condMean, sigma = tmp$condVar)

  ret <- matrix(NA, ncol = m, nrow = n_samples)
  ret[, index_given] <- rep(x_test_gaussian, each = n_samples)
  ret[, dependent_ind] <- ret0

  colnames(ret) <- cnms
  return(as.data.table(ret))
}
```

What takes time here is the call to `condMVNorm::condMVN`, the isSymmetric-test and `mvnfast::rmvn`, and this is done for every test observation and every feature combination.

What should/could be done:

1. Re-use the conditional covariance matrices and the check that these are symmteric. These are unnecessarily computed multiple times although they are the same regardless of the test observation. 
2. Sample once per feature combination and only add the test-observation-dependent mean in a secondary call.
3. Potentially do sampling only once and convert the samples using the Cholesky decomposition of the covariance matrix and adding the mean [this is not necessarily faster]

LHBO:
In this pull request, we improve the efficiency of the Gaussian and (Gaussian) copula methods.

The efficiency has been improved by
1. Computing the conditional Gaussian distributions only once and then adding the explicand-specific parts to the precomputed quantities as outlined in #231.
2. Sampling MC samples only once from a standard normal and then transforming them to the correct conditional normal distribution.
3. Rewriting the computationally heavy parts of the code into C++ code to speed up the computations.

In this pull request, the following C++ functions replace the equivalent R functions:
- The C++ function `prepare_data_gaussian_cpp` replaces the R function `prepare_data_gaussian`.
- The C++ function `prepare_data_copula_cpp` replaces the R function `prepare_data_copula`.
- The C++ function `inv_gaussian_transform_cpp` replaces the R function `inv_gaussian_transform`.
- The C++ function `quantile_type7_cpp` replaces the R function `quantile(..., type = 7)`.

Closes #231


martinju:
This is great! I am most in favor v5, which I added a simplification to, using rnorm directly instead of rmvnorm. This is slightly faster. I didn't check, but I think one can then also go directly to the transpose of B so that may help even more. 
Further, since we are improving code here, it would be nice to see whether the whole double-lapply think can be done with 2 basic for loops in C++. All of that code is essentially just matrix multiplication etc which is much fast in C++. It is quite much code to transform, but ChatGPT should be super-helpful here, and you can also see the other cpp-files in shapr for how to set it up.  Note that it does not matter if the c++-code is not "efficient" using loops and so on, it will anyway be fast since it is being complied. 

Feel free to try that out if you want. Otherwise, I can give it a go next week.

LHBO:
> This is great! I am most in favor v5, which I added a simplification to, using rnorm directly instead of rmvnorm. This is slightly faster. I didn't check, but I think one can then also go directly to the transpose of B so that may help even more. Further, since we are improving code here, it would be nice to see whether the whole double-lapply think can be done with 2 basic for loops in C++. All of that code is essentially just matrix multiplication etc which is much fast in C++. It is quite much code to transform, but ChatGPT should be super-helpful here, and you can also see the other cpp-files in shapr for how to set it up. Note that it does not matter if the c++-code is not "efficient" using loops and so on, it will anyway be fast since it is being complied.
> 
> Feel free to try that out if you want. Otherwise, I can give it a go next week.

I will look at it next week. I agree that using C++ will be beneficial

src/Copula.cpp
```diff
@@ -0,0 +1,229 @@
...
+      qs(i) = (1 - h) * qs(i) + h * sorted_x(hi(i) - 1);
+    }
+  }
+
+  return qs;
+}
+
+//' Transforms new data to a standardized normal distribution
+//'
+//' @details The function uses `arma::quantile(...)` which corresponds to R's `stats::quantile(..., type = 5)`.
```
martinju, line 47: Correct this. Uses type = 7.

LHBO: Removed the whole sentence as we are not longer using arma::quantile()

LHBO:
> @Lars: All good except the manual update mentioned above, and the ties-issue mentioned on slack. I will approve the failing tests once that is fixed, and you have confirmed that the tests differences are due to sampling error only.

- Fixed the manual update mentioned above. 
- I fixed the ties issues by using the R rank function for that problem. We do not need C++ code there, as the code takes milliseconds to run.
- I added a code example in `shapr/inst/scripts/compare_copula_in_R_and_C++.R` where I compare the Shapley values produced by the R and C++ versions. The difference tends to zero when `n_samples` tends to infinity. For `n_samples = 1000000`, the max absolute difference is in the third decimal point.

martinju (approved):

PR #367, opened 2023-12-04
Include paired combination sampling

LHBO:
In this pull request, we add support for including paired combination sampling as an option when using a sampled subset of combinations to compute the Shapley value explanations. The user can specify whether he/she wants to use `unique` (default before, and still default), `unique-paired` (new), or `non-unique` (default in very old version of `shapr`).

Paired sampling has been shown to improve the precision of the approximated Shapley value explanations in several papers., see e.g., (https://arxiv.org/abs/2107.07436). 

In my own simulation studies, I found it beneficial to use paired sampling when the features were correlated, but I saw no improvement for independent features.


R/explain.R
```diff
@@ -33,6 +33,22 @@
 #' If `n_combinations = NULL`, the exact method is used and all combinations are considered.
 #' The maximum number of combinations equals `2^m`, where `m` is the number of features.
 #'
+#' #' @param combination_sampling_method Character.
+#' The sampling method used to sample the combinations when `n_combinations` is specified, i.e., not `NULL`.
+#' It should either be `"unique"` (default), `"unique-paired"`, or `"non-unique".`
```
LHBO, line 38: Dot inside

PR #377, opened 2024-02-19
Update the imports

LHBO:
In this pull request, we remove imports that are no longer needed in the `shapr-package.R` file. These imports were important at an earlier version of the `shapr` package, but were not removed when the corresponding dependencies were removed.


LHBO:
Note that several more can be removed if one use e.g. `data.table::copy` instead of just `copy`. Currently there is no consistency across methods when to include the library prefix and not. Discuss with Martin.

Also, `sourceCpp` is no longer needed, but removing that makes the checks give a lot of warnings.

LHBO:
Should remove whitespace around the number of features.

`Error in check_approach(internal) : 
  `approach` must be one of the following: 
 categorical, copula, ctree, empirical, gaussian, independence, regression_separate, timeseries, vaeac 
 or a vector of length one less than the number of features ( 3 ), with only the above strings.`

martinju:
Do this in #391

PR #381, opened 2024-03-05
Make vaeac vignette visible on pkgdown site

martinju:
+ modify vignette menu title of main vignette


LHBO:
I agree with the new titles and good that you made the vaeac vignette visible.

LHBO (approved):
I agree with the new titles and good that you made the vaeac vignette visible.

PR #384, opened 2024-03-25
Enabling the MSEv criterion for the Python wrapper

LHBO:
The MSEv criterion was not returned in the Python wrapper. I fixed this and updated the examples to accept it as an output from the `explain` function.


martinju:
The failing tests are unrelated to anything in this PR (newer R versions on some systems complain about minor bugs in the documentation), so I will merge the PR anyway.

PR #389, opened 2024-04-11
Fix R devel notes

martinju:
Recent updates to R-devel complains about some bugs in our documentation. The PR fixes these bugs


R/approach_vaeac.R
```diff
@@ -781,7 +781,7 @@ vaeac_train_model_auxiliary <- function(vaeac_model,
       train_vlb = as.array(train_vlb$cpu()),
       val_iwae = as.array(val_iwae$cpu()),
       val_iwae_running = as.array(val_iwae_running$cpu()),
-      parameters = last_state
```
LHBO, line 784: Would prefer it to be the way it was before, but rather change `last_epoch` to `last_state` in the one line in the roxygen of `vaeac_print_train_summary`. Then no other changes are needed

man/vaeac_print_train_summary.Rd
```diff
@@ -4,7 +4,7 @@
 \alias{vaeac_print_train_summary}
 \title{Function to printout a training summary for the \code{vaeac} model}
 \usage{
-vaeac_print_train_summary(best_epoch, best_epoch_running, last_state)
+vaeac_print_train_summary(best_epoch, best_epoch_running, last_epoch)
```
LHBO, line 7: Rebuild manual after changing back to `last_state` in the function call and rather updating the oxygen @param.

LHBO (changes requested):
Rather change to `last_state` in the roxygen documentation, as that is where the logical error was.

R/approach_vaeac_torch_modules.R
```diff
@@ -2064,7 +2060,7 @@ specified_masks_mask_generator <- torch::nn_module(
     if (!is.null(seed)) set.seed(seed) # Set seed if the user specifies a seed for reproducibility.
     nan_mask <- batch$isnan()$to(torch::torch_float()) # Check for missing values in the batch
     n_masks <- nrow(masks) # Get the number of masks to choose from
-    size <- nrow(batch) # Get the number of observations in the batch
+    size <- nrow(batch) # Get the number of observations in the batch+
```
martinju, line 2063:
```suggestion
    size <- nrow(batch) # Get the number of observations in the batch
```

LHBO (approved):
Looks good!

PR #390, opened 2024-04-11
Not require torch

martinju:
Currently, installation of shapr fails if torch is not installed, even though torch is only needed for the vaeac approach and included as a suggested package. 

The code in this PR is tested and works in a local environment which does not have torch installed.


martinju:
@LHBO Please look at #389 first as this is built directly on top of that so it includes that code as well. The "changed files" should reduce to just one file one that is merged. Sorry about that.

LHBO:
> @LHBO Please look at #389 first as this is built directly on top of that so it includes that code as well. The "changed files" should reduce to just one file one that is merged. Sorry about that.

The solution of wrapping the file in `requireNamespace` seems fine to me. However, need to incorporate the requested changes from #389 before merging.

LHBO (changes requested):
> @LHBO Please look at #389 first as this is built directly on top of that so it includes that code as well. The "changed files" should reduce to just one file one that is merged. Sorry about that.

The solution of wrapping the file in `requireNamespace` seems fine to me. However, need to incorporate the requested changes from #389 before merging.

LHBO:
Should one in this PR also add the following check to `vaeac` such that it works without progressr? It did not work from Python or something you mentioned at some time, but I have forgotten the exact details. 
```  
if (requireNamespace("progressr", quietly = TRUE)) {
    p <- progressr::progressor(...)
  } else {
    p <- NULL
  }
  ```

martinju:
I am still not able to make this work out. What I have tried:

1. Wrap everything in if(requireNamespace("torch")){ FUNCTIONS + roxygen} 
- This was worked out but gave errors on the form: "Block must have a @name.
ℹ Either document an existing object or manually specify with @name" when calling devtools::document()
2. Wrap every single torch-module in FUNCTION_NAME <- ifelse0(!requireNamespace("torch"),"nothing","FUNCTION") with the roxygen commands outside of this. 
- This gave wierd-looking documentation with no input for the arguments
3. As suggested by chatGPT, I added a placeholder for the function in the main .R-function file and then made a conditional re-definition of that in the .onLoad function under zzz.R. 
- This also gave wierd documentation. 

To show the issues, in the pushed code memory_layer() is defined without conditioning, skip_connection is defined as in point 3, and the rest is as in point 2.  

I don't know how to proceed with this for now. 

One (unsatisfactory) last way out of this is perhaps to delete all documentation for these objects? 

@LHBO Feel free to try to solve it. If we don't succeed, we may try to create a small reproducable example package with this issue and bring it on stack overflow or similar...

LHBO:
This was fixed in #393 and we therefore close this PR.

PR #393, opened 2024-04-17
Restructured torch modules to support shapr installation without torch

LHBO:
In this PR, we refactor the `vaeac` approach so that the `torch`-modules are initiated through functions. Previously, when they were not inside functions, installation of `shapr` failed as it tried to evaluate `torch::`, which was not installed. 
We tried to fix this in #390, but do to technical issues for me, I had to make a new PR. 

Also fixed some typos in the Roxygen documentation and ensured that the `progressr` progress bar inside the `vaeac` approach is only called if `progressr` is available.

Details:
Instead of having
```
vaeac_dataset <-  torch::dataset(
  name = "vaeac_dataset", 
  initialize = function(X, one_hot_max_sizes) {...},
  .getbatch = function(index) {...},
  .length = function() {...}
 )
```
we replaced it with
```
vaeac_dataset <- function(X, one_hot_max_sizes) {
  vaeac_dataset_tmp <- torch::dataset(
    name = "vaeac_dataset", 
    initialize = function(X, one_hot_max_sizes) {...},
    .getbatch = function(index) {...},
    .length = function() {...}
  )
  return(vaeac_dataset_tmp(X = X, one_hot_max_sizes = one_hot_max_sizes))
}
```
Some extra care had to be given to `memory_layer`, which had an internal and shared environment between all instances of `memory_layer`. In the new version, we have to create an environment first and then send this to the new version of `memory_layer`.


martinju (approved):
I tested this locally both with and without the torch package, and it seems to work fine.

To test it installs fine without the torch package, deleted the torch package and then installed the pacakge form this branch by `devtools::install_github("LHBO/shapr",ref = "Lars/torch_module_restructure")`. I can then run e.g. the readme example. When doing devtools::document() I get errors that torch is not installed. This seems to be related to the  reference to torch in the documentation. That is anyway fine, and similar to what we get trying to run an example involving torch.

When having torch installed, all checks pass locally, and I can run devtools::document() without errors.

I just checked the code briefly to understand conceptually what you did, but didn't inspect it in detail. 
Please write a brief summary of what you did in the PR description, then you can just merge.

Well done!

PR #398, opened 2024-09-10
Fix cutting of coalition list per horizon in ```shapley_setup_forecast```

jonlachmann:
This makes the tests for forecast explanation run, and some of them work. There are however, still ones that do not work.


martinju:
Hi

I did some mine edits in convergence, and then merged these changes into that to see if some of the forecast-tests then got fixed. A few of them did. I just pushed the changes to the convergence branch, so you may want to get the latest version of that before continuing on this PR.

What remains to fix in this PR thus seems to be the following tests in forecast-output:

L60: "forecast_output_forecast_ARIMA_group_numeric"
L153: "ARIMA gives the same output with different horizons with grouping" +

Which both have to do with grouping. I suspect something odd happened when I changed the naming conventions in X as described by email. At least I would recommend starting there.

PR #396, opened 2024-06-07
Restructure explain() for iterative estimation with convergence detection, verbose arguments ++

martinju:
Very early draft. Lots of cleanup and moving things around remains, but the general overall structure will probably be close to what we got here.

To be done in this PR (some may be removed here and handled in separate PRs):

- [x] Add iterative kernelshap estimation: Sampling some coalitions at a time and estimating shapley values. Iterate until convergence threshold is met (based on estimated sd).
  - [X] for features
  - [x] for groups
- [x] Add bootstrapping approach estimating for standard deviation
  - [X] for features
  - [x] for groups
- [x] Add paired sampling of coalitions
  - [X] for features
  - [x] for groups
- [X] Add reweighting of sampling shapley kernel weights
- [X] Restructure non-iterative (classical) approach as a special case of the iterative approach for code simplicity
- [x] Rewrite batch_computing to be computed within each iteration 
- [x] Remove all traces of iterative arguments put outside of the iter_list
- [x] Consider moving form n_batches to something like max_batch_size + min_n_batches as input argument. The former controls the memory allocation (smaller means less memory consumption), while the latter ensures we get at least some progress updates while doing the updating. An internal function could then set the number of batches at each iteration to min(n_combinations,max(min_n_batches,ceiling(n_combinations/max_batch_size))))
- [x] Remove the stop() calls based on the number of features. 
- [ ] Add parallelization of the bootstrapping function
- [x] Check that the new code structure works with parallelization (via future.lapply)
- [x] Add argument `verbose` with arguments - Add verbose = c("basic","shapley","vS_details"), with "basic" as the default, showing what is currently going on in the function, the filename of the tempfile, and what iteration we are at (+ later estimate of the remaining computationt time) NULL or "" should give no printout at all, "shapley" means printing intermediate shapley estimates, "vs_details" means printing results while estimating the vS_functions (where this is done in more than a single step).
- [x] add save_intermediate = c("to_disk","in_memory") to allow the user so save results to file while doing the estimation. May choose to save everything to disk or just what is needed to continue adding coalitions if we are not yet happy with the results.
- [x] Add tests for the iterative approach
- [x] Set new defaults in explain(): paired_sampling =TRUE, shapley_reweighting = "on_N" ++
- [x] Carefully update all test checking files in at least 3 steps: 1. ensure classical shapley estimates gives the same answer (keeping the same order of setup_approach and shapley_setup), 2. update shapley values with the new ordering, 3. update the internal objects
[...]

PR #400, opened 2024-10-04
Asymmetric causal Shapley values with adaptive sampling

LHBO:
Extension of PR #395 that builds on the adaptive sampling introduced in PR #396. As the latter PR completely rewrote all of the main functions in shapr, I found it easier to start on a new one as there were a lot of merge conflicts when trying to merge #396 into #395.

In this PR, we add support for computing asymmetric and/or causal Shapley values. The asymmetric version can use all approaches, while the causal version is limited to the Monte Carlo-based approaches. The implementation is an extension of #273 (but this PR was restricted to the `gaussian` approach and the old version of `shapr`), which was adapted from the package [CauSHAPley](https://gitlab.science.ru.nl/gbucur/caushapley/).

Asymmetric Shapley values were proposed by [Frye et al. (2020)](https://proceedings.nips.cc/paper/2020/file/0d770c496aa3da6d2c3f2bd19e7b9d6b-Paper.pdf) as a way to incorporate causal knowledge in the real world by restricting the possible permutations of the features when computing the Shapley values to those consistent with a (partial) causal ordering.

Causal Shapley values were proposed by [Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf) as a way to explain the total effect of features on the prediction, taking into account their causal relationships, by adapting the sampling procedure in shapr.

The two ideas can be combined to obtain asymmetric causal Shapley values. If you would like more details, you can see Heskes et al. (2020).

**Usage:**  (Assume `N_features = 7`)
_(Symmetric) Conditional Shapley values_: `asymmetric = FALSE` (default), `causal_ordering = list(1:7)` (default), and `confounding = FALSE` (default)

_Marginal Shapley values_: either 1) the same as above, but set approach = independence, or 2) `asymmetric = FALSE` (default), `causal_ordering = list(1:7)` (default), and `confounding = TRUE`.

_Asymmetric conditional Shapley values_ with respect to a specific ordering: `asymmetric = TRUE`, `causal_ordering = list(1, c(2, 3), 4:7)`, and `confounding = FALSE` (default).

_Causal Shapley values_ (compute all coalitions, but chains of sampling steps): `asymmetric = FALSE` (default), `causal_ordering = list(1, c(2, 3), 4:7)`,  and`confounding = c(FALSE, TRUE, FALSE)`.

_Asymmetric Causal Shapley values_ (compute only coalitions respecting the ordering and chains of sampling steps): `asymmetric = TRUE`, `causal_ordering = list(1, c(2, 3), 4:7)`, and `confounding = c(FALSE, TRUE, FALSE)`.

**Main differences:**
The user now has the option to specify  `asymmetric`, `causal_ordering`, and `confounding` in the `explain` function.
The first argument, `asymmetric`, specifies if we should consider all feature combinations/coalitions, or only the combinations that respect the (partial) causal ordering given by `causal_ordering`. The second argument, `causal_ordering` is a list specifying the (partial) causal ordering of the features (groups), i.e., `causal_ordering = list(1:3, 4:5)`, which implies that features one to three are the ancestors of four and five. The third argument, `confounding` specifies if the user assumes that each component is subject to confounding or not, e.g., `causal_confounding = c(FALSE, TRUE)`. So that you know, practitioners are responsible for correctly identifying the causal structures.

When the `causal_ordering` is **not** `list(1:N_features)`, then we have a causal structure that implies that some coalitions/feature combinations will not respect the order. For example, we cannot have a combination that conditions/includes feature four and not all of the features one to three in the setting above, as they are feature four's ancestors. If `asymmetric = TRUE`, then we only use the combinations that respect the order. If `asymmetric = FALSE`, then we use all combinations. Furthermore, generating the MC samples for each valid coalition will introduce a chain of sampling steps, which the `confounding` argument influences. 

That is, if `S = {2}`, we would in the first step (assuming `confounding = c(FALSE, TRUE)`) sample `X1, X3 | X2`, and in the second step, we would sample `X4, X5 | X1, X2, X3`. The confounding changes whether to include the features in the same component as conditional features or not, as Heskes et al. (2020) explained. Also, see examples in `get_S_causal()` for demonstrations of how changing the confounding assumption changes the data generation steps.

To reuse most of the `shapr` code, we iteratively call `prepare_data()` with different values of `S` to generate the data. This introduces a lot of redundant computations, as we then generate `X1, X3, X4, X5 | X2` in the first step, but throw away `X4` and `X5`. To only generate MC samples for the relevant features, we would have to rewrite all  `prepare_data.approach` functions also to take in a Sbar argument as they currently assume that Sbar is all features not in S. 

The `independence,` `empirical`, and `ctree` approaches can not necessarily generate `n_samples` but rather weigh the samples. Combining these weights in an interactive sampling process is not obvious. We solve it by sampling the samples `n_samples` time using the weights. This means that we will have duplicates, which introduces extra computations. 

**Plot:**
Additionally, we have introduced the `include_group_feature_means = FALSE` argument in `plot.shapr` and `plot_SV_several_approaches` as some plots need to have a feature value, which we do not have for group-wise Shapley values. When `TRUE`, we use the average feature value among the features in each group (for each explicand).
In `plot_SV_several_approaches`, we also add the `index_explicands_sort` argument to decide the order of the explicands in the plots.  

TODO:
- [x] Implement exact asymmetric causal feature Shapley values for all Monte Carlo-based approaches.
- [x] Implement support for non-exact. Need to figure out how to sample the allowed combinations and what weights to give them. I have a function that can create all valid combinations but still grows at `O(2^C)`, where C is the number of features in the largest component in the causal ordering.
- [x] Implement support for groups. Restrict feature groups to be in the same component in the causal ordering. 
- [x] Ctree is not very fast due to many inputs/MC samples. Can we somehow use the weights to speed it up?
- [x] Create a small vignette. 
- [x] Make vignette runnable.
- [x] Add new tests that test that it works both in the adaptive and regular version

FUTURE:
- [ ] FUTURE: Generate only the features in Sbar and not all features not in S (Since Sbar union S is not all features). To do this, all `prepare_data.approach` functions must be rewritten. 
- [ ] FUTURE: Be more clever in choosing the combinations to go into the different batches. Combinations that condition on features in the same components have similar chains of sampling steps. Often, only the first step is different. See `?get_S_causal` for some examples of chains. E.g., we can have `c("2|", "3,4|1,2", "5|1,2,3,4")` and `c("1|", "3,4|1,2", "5|1,2,3,4")`. Could then ideally save some time by computing the rest together to minimize the number of times we have to recompute/model the same conditional distributions (the last step is often identical for all combinations).
- [ ] FUTURE: The causal versions of Gaussian and copula can be written faster in C++ by sending the whole chain of sampling steps for each coalition, but then we would no longer have the same structure for all sampling methods.

References:
- Heskes, T., Sijben, E., Bucur, I. G., & Claassen, T. (2020). Causal Shapley Values: Exploiting Causal Knowledge to Explain Individual Predictions of Complex Models. Advances in Neural Information Processing Systems, 33.
- Frye, C., Rowat, C., & Feige, I. (2020). Asymmetric Shapley values: incorporating causal knowledge into model-agnostic explainability. Advances in Neural Information Processing Systems, 33.


LHBO:
Ready to merge into #402 when discussed with @martinju

R/explain.R
martinju, line 1: Please update using the format used for the other input arguments in explain. First line is just what kind of input it takes (list/string/logical). The next sentence should in these case be whether it is applicable for (regular) non-causal/asymmetric explanations. Since 99% of our users probably want to symmetric and non-causal we should try to confuse them as little as possible with these extra arguments. We could also consider putting them into a list of extra calls, but lets keep them like they are for now at least (just clarifying the documentation)

R/finalize_explanation.R
```diff
@@ -248,6 +248,10 @@ finalize_explanation_forecast <- function(vS_list, internal) { # Temporary used
   internal$main_timing_list <- internal$iter_timing_list <- internal$timing_list <- NULL
 
 
+  # Clearing out the timing lists as they are added to the output separately
+  internal$main_timing_list <- internal$iter_timing_list <- internal$timing_list <- NULL
```
martinju, line 252: already added. delete

martinju (changes requested):
I have looked at most of the new stuff. I have not looked at setup yet, hope to be able to do this soon, so we can wrap this up. See the other comments.

R/approach_vaeac.R
```diff
@@ -76,7 +73,18 @@ setup_approach.vaeac <- function(internal, # add default values for vaeac here.
   parameters <- c(parameters[(length(vaeac_main_para) + 1):length(parameters)], parameters[seq_along(vaeac_main_para)])
 
   # Check if vaeac is to be applied on a subset of coalitions.
-  if (!parameters$exact || parameters$is_groupwise || combined_approaches) {
+  if (!is.null(S_causal)) {
```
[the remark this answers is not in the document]
LHBO: Added this

R/compute_vS.R
```diff
@@ -169,8 +169,10 @@ batch_prepare_vS_MC <- function(S, internal, model, predict_model) {
   y <- internal$data$y
   xreg <- internal$data$xreg
   keep_samp_for_vS <- internal$parameters$keep_samp_for_vS
+  causal_sampling <- internal$parameters$causal_sampling
 
-  dt <- batch_prepare_vS_MC_auxiliary(S = S, internal = internal) # Make it optional to store and return the dt_list
+  # Make it optional to store and return the dt_list
+  dt <- batch_prepare_vS_MC_aux_causal(S = S, internal = internal, causal_sampling = causal_sampling)
```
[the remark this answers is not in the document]
LHBO: Ranamed batch_prepare_vS_MC_aux_causal  to batch_prepare_vS_MC_auxiliary

R/shapley_setup.R
```diff
@@ -139,20 +171,31 @@ shapley_setup <- function(internal) {
...
                                    approach0 = "gaussian",
-                                   shapley_reweighting = "none") {
+                                   shapley_reweighting = "none",
+                                   dt_valid_causal_coalitions = NULL) {
   if (exact) {
-    dt <- exact_coalition_table(m, weight_zero_m)
+    dt <- exact_coalition_table(
+      m = m,
+      weight_zero_m = weight_zero_m,
+      dt_valid_causal_coalitions = dt_valid_causal_coalitions
```
martinju, line 188: not superimportant, but we could maybe call this dt_valid_coalitions, just add a comment on this line and in the documentation that it is only relevant for for causal?

martinju: same for the sampling version

R/explain.R
```diff
@@ -138,6 +139,37 @@
...
+#' Frye et al. (2020). If `confounding` is provided, i.e., not `NULL`, then `shapr` computes asymmetric causal
+#' Shapley values as specified in Heskes et al. (2020).
+#'
+#' @param causal_ordering List of vectors specifying the components of the partial causal ordering.
+#' The features in each component are given by the feature/group names or their indices.
+#' If `causal_ordering` is `NULL` (default), then no causal ordering is assumed, i.e., a causal ordering with
+#' a single component containing all features `list(1:n_features)` or groups `list(1:n_groups)` for
+#' feature-wise and group-wise Shapley values, respectively. If the user specifies the features/groups
+#' by their names (strings), then `shapr` will convert them to the feature/group indices specified by the
+#' order in `get_model_specs` for features and `groups` for groups. For feature-wise Shapley values,
```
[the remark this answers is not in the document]
LHBO: Removed and fixed

R/explain.R
LHBO: Updated

R/shapley_setup.R
```diff
@@ -331,7 +404,20 @@ sample_coalition_table <- function(m,
   X[, N := 1]
   ind <- X[, .I[data.table::between(coalition_size, 1, m - 1)]]
   X[ind, p := p[coalition_size]]
-  X[ind, N := n[coalition_size]]
+
+  # TODO: Discuss with Martin. Should N be
+  #  1. the total number of coalitions of a size
+  #  2. or only the number of valid coalitions of said size (This verson is now implemented below)
```
martinju, line 410: this is what we discussed right? can remove the todo in that case

inst/extdata/day.csv
martinju, line 1: do we need this extdata? remove the entire folder if not.

martinju (changes requested):
Done with the rest of the review. See the last minor comments. Not crucial to change the name.

R/approach_copula.R
```diff
@@ -57,18 +57,41 @@ prepare_data.copula <- function(internal, index_features, ...) {
   copula.mu <- internal$parameters$copula.mu
   copula.cov_mat <- internal$parameters$copula.cov_mat
   copula.x_explain_gaussian_mat <- as.matrix(internal$data$copula.x_explain_gaussian)
+  causal_sampling <- internal$parameters$causal_sampling
+
+  # Update the number of MC samples for causal Shapley values not in the first step
+  causal_first_step <- isTRUE(internal$parameters$causal_first_step) # Only set when called from `prepdare_data_causal`
+  n_MC_samples_updated <- if (causal_sampling && !causal_first_step) n_explain else n_MC_samples
+
+  # Update the `copula.x_explain_gaussian_mat` for causal Shapley values not in the first step
```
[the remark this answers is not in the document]
LHBO: Fixed

R/approach_gaussian.R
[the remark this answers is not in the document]
LHBO: FIxed

inst/extdata/day.csv
LHBO: Keep it after discussing it. This bike rental data set is used in the Vignette.

R/shapley_setup.R
LHBO: I discussed it with Martin, and we decided to let it be like this for now.

martinju (approved):
Let's say this is good now :-)

martinju:
We're aware that the mac-os test fails for the categorical approach, where one observation gets a different shapley value, due to a change in dt_vS value of v(S={3}). Further debugging is required to figure this out, but it is not straight forward as it only happens on the GHA macOS version, not locally on mac. Would probably need to add a new test which with keep_vS_output = TRUE to figure out more precisely where this is occuring.

PR #406, opened 2024-10-12
Update the `cex` argument in `ggbeeswarm::geom_beeswarm()`

LHBO:
In the current version, the `cex` argument in `ggbeeswarm::geom_beeswarm` does not scale well when `N_explain` increases as it is set to `0.4`. In this PR, we change it to `cex = 1 / length(index_x_explain)^(1/4)`.

Another option is to use `corral` as illustrated in `inst/scripts/Beeswarm_illustration.R`.  This file compares the old method, the new version with the modified cex argument, another version where we use coral, and the code used in https://arxiv.org/abs/2410.04883. Here we let `N_explain` be 50, 100, 1000, and 5000, and the new version with the modified cex argument provides nice plots for all settings and most similar to what the old code did for small values of `N_explain` but now also for larger values.


LHBO:
I have not accepted the changes in plot tests.

PR #421, opened 2024-11-22
Update vignettes ++

martinju:
- Going through all vignettes + readme, fixing various issues with cache, consistency, style, code and text
- Fix bug in copula for asymmetric SV
- Fix bug for causal with groups


martinju:
GHA fails are due to mac and lintr errors (both unrelated to shapr)

PR #425, opened 2024-12-11
Bug in `explain_forecast()`

LHBO:
In this PR, we fix 2 main bugs in `explain_forecast()`:

A: Using `verbose = "shapley"`, caused the function to crash.
B: Inconsistencies related to the number of coaltions used and shapley values computed with `explain_forecast()`.

Bug A stemmed from :
1. There was an inconsistency in that `internal$iter_list[[iter]]$dt_shapley_est` contained the index columns `"explain_idx", "horizon"` while `internal$iter_list[[iter]]$dt_shapley_sd` did not. We added the index columns to the latter.

2. When creating a data.table to print with the iterative Shapley values estimates with standard deviations, the code did not take into consideration that the number of rows is not `n_explain` when doing forecast, but rather `n_times` x `horizon`. This is now fixed by using the number of rows in `internal$iter_list[[iter]]$dt_shapley_est` instead.

3. To be consistent with regular Shapley value, we remove `"explain_idx", "horizon"` when printing the iterative Shapley value estimates and standard deviation. In regular Shapley we omit `"explain_idx"`. 
Note. It might be a good thing to include `"explain_idx"` (and `"horizon"`), but then the print function needs to be rewritten to reflect that `"explain_idx"` (and `"horizon"`) is not an estimate with a standard deviation.
Note 2. There is an inconsistency between `explain` and `explain_forecast`, where the former use `explain_id`, while the latter use `explain_idx`. 

Bug B was fixed by defining the number of coaltions to be the number of coalitions used per horizon. We also introduced a new parameter `n_sampled_coalitions` which for explain_forecast() is the number of coalitions per horizon -2 (zero and full prediction), and for regular explain() it is the number of coalitions. A number of additional minor fixed related to the number of shapley values computed with/without grouping was also added. As a side-effect this also fixed the incorrect verbose printout for `explain_forecast()`.

PS. 
`print_dt <- as.data.table(matrix(paste(matrix1, " (", matrix2, ") ", sep = ""), nrow = nrow(matrix1)))`
produce extra whitespace at the end. Should change to, e.g.,
`print_dt <- as.data.table(matrix(paste0(matrix1, " (", matrix2, ")"), nrow = nrow(matrix1)))`.


LHBO:
To see that bug no longer occurs, run:
 ```
 devtools::load_all()
explain_forecast(
  testing = TRUE,
  model = model_arima_temp,
  y = data_arima[1:150, "Temp"],
  xreg = data_arima[, "Wind"],
  train_idx = 3:148,
  explain_idx = 149:150,
  explain_y_lags = 3,
  explain_xreg_lags = 3,
  horizon = 3,
  approach = "empirical",
  phi0 = p0_ar,
  group_lags = FALSE,
  max_n_coalitions = 150,
  iterative = TRUE,
  iterative_args = list(initial_n_coalitions = 10),
  verbose = c("basic", "progress", "convergence", "shapley", "vS_details")
)
 ```

LHBO:
I am confident that the test that fails now are actually correct, in the sense that the output of the old runs were incorrect.
As in the previous code in `check_convergence.R`, the forecast `dt_shapley_sd` did not contain the index columns and then  `max_sd <- dt_shapley_sd[, max(.SD, na.rm = TRUE), .SDcols = -1, by = .I]$V1` removed `none` column, while `dt_shapley_est0[, maxval := max(.SD, na.rm = TRUE), .SDcols = -c(1, 2), by = .I]` removed `explain_idx` and `horizon` but kept `none`, as `dt_shapley_est` had the index columns.

martinju (approved):

PR #432, opened 2025-01-07
Fix of failing code line in README

osorensen:
The example fails since R does not recognize the function `handlers()`.


martinju (approved):

martinju:
Thanks! @osorensen

PR #435, opened 2025-01-20
Fix cpp pointer bug

martinju:
Fix CRAN NOTE which turned out to be bug related to returning NULL rather than integer(0) to identify (unconditional)
asymmetric causal sampling: updating the `get_S_causal_steps` function to handle empty intersections.

Also adding a new test for checking equality of NULL confounding with regular symmetric conditional shapley values.


martinju:
@LHBO Do you mind taking a 30 sec look at R/asymmetric_and_casual_Shapley.R
The rest is vignette updates ++

PR #451, opened 2025-04-23
skip_if_not_installed for suggested packages

martinju:
* Introduced `skip_if_not_installed` for various tests to ensure they are skipped gracefully when dependencies are unavailable. 
* Various other minor fixes


R/setup.R
```diff
@@ -228,16 +228,6 @@ get_parameters <- function(approach,
 
   # verbose
   check_verbose(verbose)
-  if (!is.null(verbose) &&
-    (!is.character(verbose) || !(all(verbose %in% c("basic", "progress", "convergence", "shapley", "vS_details"))))
-  ) {
-    stop(
```
LHBO, line 234: Any reason why removing this check?

PR #458, opened 2025-06-10
Julie/sage

julienbj:
- Computation of SAGE-values, including optional user-input of loss-function
- Waterfall- and bar- plotting of SAGE-values
- Name changes of variables ("shap" to "shapley")
- Tests for SAGE implementation and associated parameters


martinju:
SAGE implemented in #503 (largely based on this work)

PR #474, opened 2025-09-23
Asymmetric Shapley values

LHBO:
Started investigating the asymmetric problems.

Implemented a function `weight_matrix_permutation()` that takes in the causal order and outputs the weight matrix `W` which we should have used to compute the Shapley values in the asymmetric setting. It might contain some edge case logical errors, but it gives what I got by hand for some small scenarios.

Compared how the Shapley values look for the Bike dataset when we use this new weight matrix, and we got some minor changes in the Shapley values, but no huge changes.

### DIFF
--- a/inst/devel/ASV_bug.R
+++ b/inst/devel/ASV_bug.R
@@ -1,0 +1,600 @@
+rm(list = ls())
+
+# Libraries -------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


martinju:
Closing to clean up. We are in good state with asym now, but might revisit this later.

PR #496, opened 2026-06-10
Fix cli timing bug due to wrong environment

LHBO:
Previously, there was a bug that `cli` printed out that `"Computing v(S) [8ms]"` only took a couple of ms independently off how long it actually took.

The logical error is that `cli::cli_progress_step` ties the progress bar's lifetime to the calling function's environment. Since it's called inside `cli_compute_vS()`, the step is marked "done" when `cli_compute_vS()` returns, which is before the actual `v(S)` computation runs. That's why it shows 8ms.

The fix is to pass `.envir = parent.frame()` to `cli::cli_progress_step` so the progress bar's lifetime is tied to `compute_vS()` instead of `cli_compute_vS()`.

The other `cli` calls happens inside the correct function environment.

Now we get a correct printout:
`v Computing v(S) [3.4s] `

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@
 
 ### Bug fixes
 * Fixed handling of ordered factor features by treating them as factors in feature specifications and adding stricter checks for malformed feature specification lists. ([#495](https://github.com/NorskRegnesentral/shapr/pull/495))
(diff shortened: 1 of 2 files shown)


LHBO:
@martinju, can you take a look when you have time and update the vignettes and snapshots?

After some thinking, the timing is not included in the tests/snapshots, and neither in the vignettes, so there is no point updating them (but please correct me if I am wrong), hence this is ready to merge.

martinju (approved):

PR #500, opened 2026-06-23
Rename Python wrapper from shaprpy to pyshapr

martinju:
## Summary

Renames the Python wrapper package from `shaprpy` to `pyshapr`. This is a name-only change with no
functional impact on the computations. A thin `shaprpy` compatibility shim is retained so existing
users are not broken during a transition period.

## Changes

**Python wrapper**
- Moved source `python/src/shaprpy/` → `python/src/pyshapr/`.
- `pyproject.toml`: package name → `pyshapr`, version `0.5.0`, ruff per-file-ignore path updated.
- Added a `shaprpy` compatibility shim package that depends on `pyshapr>=0.5.0`.
- Updated examples, tests, `conftest.py`, and `README.md` to the new name. Snapshots are unchanged
  (they do not capture the package name).

**R side**
- `DESCRIPTION`: wrapper name reference `'shaprpy'` → `'pyshapr'`; dev version bumped to `1.0.8.9001`.
- Regenerated `man/*.Rd`; R console output from `summary()`/`print()` on Python-initiated explanations
  now references `pyshapr.explain()` / `pyshapr.explain_forecast()`.
- `NEWS.md` and `python/CHANGELOG.md` entries for the rename (`pyshapr 0.5.0`).
- Regression vignette references updated to `pyshapr` (prose-only; no rebuild required).

## Testing

- `ruff check` and `ruff format --check`: clean.
- `devtools::document()`: no documentation drift.
- `pytest`: 56 passed, 36 snapshots passed.

NOTE: Do not merge this before we push to PyPI


README.md
```diff
@@ -68,11 +68,11 @@ to view a version of this README with the old syntax (v0.2.2).
 
 ### Python wrapper
 
-We provide a Python wrapper (`shaprpy`) which allows explaining Python
+We provide a Python wrapper (`pyshapr`) which allows explaining Python
 models with the methodology implemented in `shapr`, directly from
 Python. The wrapper calls R internally and therefore requires an
 installation of R. See
-[here](https://norskregnesentral.github.io/shapr/shaprpy.html) for
+[here](https://norskregnesentral.github.io/shapr/pyshapr.html) for
```
LHBO, line 75: Page not found, but I guess this will be updated after you push this to master and maybe do some more steps?

martinju: Yes! Høna og egget dette her 😃

martinju: It will be automatically created when pushing this to master.

python-shim/src/shaprpy/__init__.py
```diff
@@ -0,0 +1,40 @@
+"""Compatibility shim for the renamed ``pyshapr`` package.
```
LHBO, line 1: This is very clever!

python-shim/LICENSE
```diff
@@ -0,0 +1,2 @@
+YEAR: 2019
```
LHBO, line 1: Is this the right year?

LHBO: I see the .md file also says 2019. Maybe that is the year shapr was created the first time?

python-shim/pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[build-system]
+requires = ["setuptools>=69", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "shaprpy"
+version = "0.4.4"
+description = "Compatibility shim: 'shaprpy' has been renamed to 'pyshapr'. Installs and forwards to 'pyshapr'."
+readme = "README.md"
+license = { file = "LICENSE.md" }
```
LHBO, line 10: Lots of extra whitespaces in the curly brackets throughout this file?

python-shim/README.md
```diff
@@ -0,0 +1,34 @@
+# shaprpy (renamed to pyshapr)
+
+> **This package has been renamed to [`pyshapr`](https://pypi.org/project/pyshapr/).**
+
+`shaprpy` is now a thin compatibility shim. Installing it pulls in `pyshapr` and
+forwards all imports to it, so existing code that does `import shaprpy` keeps
+working. Importing `shaprpy` emits a `DeprecationWarning`.
+
+This shim is maintained only for a transition period and contains no
```
LHBO, line 9: Should one specify an end date? Maybe you do it later in this branch and I have not come there yet.

python/tests/README.md
```diff
@@ -1,6 +1,6 @@
-# shaprpy Test Suite
+# pyshapr Test Suite
 
-This directory contains the test suite for shaprpy, organized as follows:
+This directory contains the test suite for pyshapr, organized as follows:
```
LHBO: To make it into a code name? This is done in the changelod.md below at least.
```suggestion
This directory contains the test suite for `pyshapr`, organized as follows:
```

python/CHANGELOG.md
```diff
@@ -1,20 +1,28 @@
...
-All notable changes to the Python `shaprpy` library will be documented in this file.
+All notable changes to the Python `pyshapr` library will be documented in this file.
 Changes to the underlying `shapr` R package is documented in
 [shapr CHANGELOG](https://norskregnesentral.github.io/shapr/news/index.html)
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.5.0]
```
LHBO, line 10: Should `shapr` and `pyshapr` have better matching version numbers? `pyshapr` is on 0.5.0 and `shapr` is 1.0.8.9001.

python/pyproject.toml
```diff
@@ -61,7 +61,7 @@ package-dir = {"" = "src"}
 
 [tool.setuptools.packages.find]
 where = ["src"]
-include = ["shaprpy*" ]
+include = ["pyshapr*" ]
```
LHBO:
```suggestion
include = ["pyshapr*"]
```

python/pyproject.toml
```diff
@@ -3,8 +3,8 @@ requires = ["setuptools>=69", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "shaprpy"
-version = "0.4.3"
+name = "pyshapr"
+version = "0.5.0"
 description = "Python wrapper for the R package shapr (via rpy2)"
 readme = "README.md"
 license = { file = "LICENSE.md" }
```
LHBO, line 10: Note done in the PR, but extra white spaces here. Similar to those mentioned above.

```suggestion
license = {file = "LICENSE.md"}
```

python/README.md
```diff
@@ -30,7 +34,7 @@ sudo apt install r-base r-base-dev -y
 
 ### 1. Install the R package `shapr`
 
-`shaprpy` requires the R package `shapr` (version 1.0.5 or newer).
+`pyshapr` requires the R package `shapr` (version 1.0.5 or newer).
```
LHBO, line 37: If the user wants to use arf, then the user needs a newer version than 1.0.5.

python/README.md
```diff
@@ -129,7 +133,39 @@ For other model types, you can supply:
...
+
+---
+
+## Supported Approaches
+
+`pyshapr` forwards all approach-specific arguments to `shapr::explain()`. Commonly used approaches include:
+
+- `"gaussian"`, `"copula"`, `"empirical"`, `"ctree"`, `"independence"`
+- `"categorical"`, `"regression_separate"`, `"regression_surrogate"`
+- `"arf"` and `"vaeac"` for mixed numerical/categorical feature sets
```
LHBO: I still think this does not make it clear which approaches you can use or mixed data

python/README.md
```diff
@@ -129,7 +133,39 @@ For other model types, you can supply:
...
+- `approach="vaeac"` requires the R package `torch` and a working torch install
+
+Minimal usage pattern:
+
+```python
+explanation = explain(
+    model=model,
+    x_train=dfx_train,
+    x_explain=dfx_test,
+    approach="arf",  # or "vaeac"
```
LHBO: Double whitespace. But looked it up: "Inline comments should be separated by at least two spaces from the statement." (PEP 8 – Style Guide for Python Code: https://peps.python.org/pep-0008/)

python/README.md
```diff
@@ -129,7 +133,39 @@ For other model types, you can supply:
...
+    phi0=dfy_train.mean().item(),
+    arf_num_trees=20,
+    arf_max_iters=5,
+    seed=1,
+)
+```
+
+For `vaeac`, pass `vaeac_*` arguments directly, for example `vaeac_epochs=10`, `vaeac_width=16`, and `vaeac_depth=2`.
```
LHBO: Same here as in the previous pr

LHBO (commented):
Only minor comments.

python/README.md
martinju: Yes, but pyshapr will work, and you can say this about every new feature being added in the future as well. While you have a point, I think it will be difficult to add a check for every change and feature in pyshapr that will require a certain version of shapr. However, what we may do is to provide a check or warning when installing or loading pyshapr along the lones of "as of the release of pyshapr version xyz (this version), shapr version abc was the most recent release. To utilize all capabilities, please make sure to use a version no older than abc". I will make a not of this to see if we may add it later.

python-shim/LICENSE
martinju: yes, so i think this should stay like this.

python-shim/pyproject.toml
martinju: Agree, well spotted. Will check.

python-shim/README.md
martinju: No, I have not done that, and was thinking of it, but was worried it may affect publishing if this is too early.

python/tests/README.md
martinju: Fair point. Will apply

python/CHANGELOG.md
martinju: I don't think that is necssary. Sometimes we do a change in shapr and not pyshapr, and the other way around, so I think that may a bit hard to maintain properly.
Since no reviewer commented on this, lets keep it flexible.

python/pyproject.toml
martinju: agree. will fix

python/README.md
martinju: Good catch. I did fix this, but apparently something got messed up during the merge to this branch, so the file now contains both the old and new version. Will fix.

python/pyproject.toml
martinju: Apparerntly this is by convention (for toml files) so will keep it like this.

PR #503, opened 2026-06-25
Adding SAGE

martinju:
## Summary

This PR adds support for **SAGE (Shapley Additive Global importancE)** values to the `shapr` package, enabling global model-loss explanations alongside per-observation Shapley values. The feature is exposed through a new `scope` argument on `explain()`, with an updated computation pipeline, SAGE-aware plotting, and new outputs surfaced through the R and Python results APIs.

## Major new feature: SAGE values

- Added SAGE support to `explain()` via a new `scope` argument (`"local"` default, `"global"` for SAGE) together with `y_explain`. When `scope = "global"`, `explain()` attributes the reduction in the global model loss (relative to always predicting `phi0`) to each feature, reusing the same machinery as regular Shapley values but replacing the value function `v(S)` with the negative expected loss `-E[loss(y, E[f(x) | x_S])]` averaged over `x_explain`. Because conditional expectations are estimated with the chosen `approach`, `shapr` can account for feature dependence, unlike the marginal sampling of Covert et al. (2020).
- The SAGE loss function is configured through `extra_computation_args$global_loss_func`, defaulting to log-loss for binary 0/1 responses and MSE otherwise, and validated by `check_extra_computation_args()`.
- Documented SAGE usage in `NEWS.md`, `python/CHANGELOG.md`, function docstrings, the `README`s, and the `general_usage` vignette, including how `scope = "global"` differs from standard (local) SHAP-style explanations.

## Computation pipeline

- `compute_shapley`, `bootstrap_shapley_inner`, and related helpers compute a single global SAGE vector using the negative model loss as the value function, adapt the bootstrap logic, and always also compute the per-observation Shapley values, storing them in `internal$output$shap_values_est`.
- User-supplied loss functions are stripped from outputs (via `testing_cleanup()`, keyed on `scope == "global"`) for reproducibility and snapshot safety.

## Results accessors (`get_results`)

- `get_results()` builds a **scope-aware** default component set and exposes new components: `scope`, `shap_values_est`, and `sage_values_est`.
- `sage_values_est` (identical to `shapley_est` for global explanations, `NULL` otherwise) is never part of the default set. `shap_values_est` is included in the default only for global (SAGE) explanations, where it carries the distinct per-observation decomposition; for local explanations it would duplicate `shapley_est` and is omitted. All components remain available on explicit request regardless of scope.

## Plotting and visualization

- `plot.shapr` and helpers support SAGE: plot types are restricted to `"bar"` and `"waterfall"`, axis labels and titles are adapted, and per-observation features are disabled for the single global explanation.

## Python wrapper

- Mirrored the R API: `explain()` takes `scope="global"` and `y_explain`; a custom Python loss is supplied via `extra_computation_args={"global_loss_func": my_loss}` and bridged into the R computation (`_build_extra_computation_args`, `_wrap_loss_func`).
- `Shapr.get_shap_values_est()` returns the per-observation Shapley values computed alongside the SAGE values, and `Shapr.to_shap()` is SAGE-aware (returns the single global loss explanation).

## Tests

- Added `test-sage-setup.R` and `test-sage-output.R` on the R side, and SAGE output/validation tests on the Python side (including a group-level SAGE output test), with matching snapshots. Summary and forecast snapshots are refreshed to record the new `scope` parameter.

---

These changes enable global loss-based explanations in `shapr` through a single, consistent `scope` argument, broadening the package's applicability and interpretability for model analysis.


.github/workflows/python-tests.yaml
```diff
@@ -44,6 +44,11 @@ jobs:
           r-version: 'release'
           use-public-rspm: true
 
+      - name: Configure R shared library path for rpy2
```
LHBO, line 47: Bit curious why these changes were needed now and not before.

python/src/pyshapr/_explain.py
```diff
@@ -103,11 +171,23 @@ def explain(
...
-      Specifies extra arguments related to the computation of the Shapley values.
+      Specifies extra arguments related to the computation of the Shapley values. When `scope = "global"`, the
+      key `"global_loss_func"` may be supplied with a Python callable taking two arguments (the true response
+      and the model prediction, in that order) and returning a single numeric loss. If omitted, logistic
+      (cross-entropy) loss is used for binary responses (values in 0/1) and mean squared error loss otherwise.
     iterative_args: dict or None, optional
       Specifies the arguments for the iterative procedure.
     output_args: dict or None, optional
       Specifies certain arguments related to the output of the function.
+    scope: str, optional
```
LHBO, line 182: I like the name. But did you find out if it supported group shapley, causal shapley, forecasting? If no, I feal like that should be mentioned here in the argument. But Sincre we have only changed the v(S), the others should work i think.

I will see later in the review if you have made a comment about it.

python/src/pyshapr/explanation.py
```diff
@@ -4,6 +4,8 @@
 
 from typing import Any
 
+import numpy as np
```
LHBO, line 7: According to PEP 8, the official Python style guide, import statements always come before from ... import statements when you are grouping your imports.
So this should come on line 4.

python/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ Changes to the underlying `shapr` R package is documented in
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
LHBO, line 10: This is to be removed later?

LHBO: I.e., "unrelased"

python/CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ Changes to the underlying `shapr` R package is documented in
...
+
+### Added
+- Added support for computing SAGE values (Shapley Additive Global importancE) via the new `explain()`
+  arguments `scope` and `y_explain`. Set `scope="global"` to compute SAGE values. A custom Python loss
+  function can be passed through `extra_computation_args={"global_loss_func": my_loss}` and is bridged into
+  the R computation. Added the `Shapr.get_shap_values_est()` accessor for the per-observation Shapley values
+  computed alongside the SAGE values, and made `Shapr.to_shap()` SAGE-aware (returns the single global loss
+  explanation). (branch: sage)
```
LHBO, line 18: Is "branch: sage" going to be a link? Otherwise I am not sure if it should be included

R/compute_estimates.R
```diff
@@ -160,6 +161,13 @@ compute_shapley <- function(internal, dt_vS) {
     }
 
     dt_kshap <- cbind(internal$parameters$output_labels, rbindlist(kshap_list, fill = TRUE))
+  } else if (sage) {
+    # SAGE: the value function is the negative model loss for each coalition, averaged over the explained
+    # observations. Applying the kernelSHAP weight matrix then yields a single set of global SAGE values.
+    vS_loss <- compute_vS_loss(dt_vS, internal)
+    kshap <- t(W %*% vS_loss)
```
LHBO: Last three lines here are identical to those in the else clause below,  if  we there define `vS_loss <- as.matrix(dt_vS[, -"id_coalition"])`. So can refactor code to save some lines of code.

R/compute_estimates.R
```diff
@@ -343,10 +377,15 @@ bootstrap_shapley_inner <- function(X,
...
 
-    kshap_boot <- t(W_boot %*% as.matrix(dt_vS[id_coalition %in% X_boot[
-      boot_id == i,
-      id_coalition
-    ], -"id_coalition"]))
+    boot_ids <- X_boot[boot_id == i, id_coalition]
+    dt_vS_boot <- dt_vS[id_coalition %in% boot_ids]
+
+    if (sage) {
+      vS_loss_boot <- compute_vS_loss(dt_vS_boot, internal)
```
LHBO: Same kind of refactoring can be done in this if else.
Just consider it.

R/explain.R
```diff
@@ -166,6 +166,20 @@
 #' The `approach` cannot be `regression_separate` or `regression_surrogate`, as the
 #' regression-based approaches are not applicable to the causal Shapley methodology.
 #'
+#' @param scope String.
```
LHBO, line 169: Good argument name

.github/workflows/python-tests.yaml
martinju: If i rememenr correctly this was a byproduct needed due the code below in this yaml -- which was introduced as an additional safety net when the R code is update while the python code isn't -- python still needs to reinstall the R package. This was not enforced before.

python/src/pyshapr/_explain.py
martinju: It suppors group (we have a test for that). forcasting is another function, so no. 
I will check causal.

python/CHANGELOG.md
martinju: yes, once released (it isn't yet)

python/CHANGELOG.md
martinju: This will be replaced with a link to this PR later.

R/plot.R
```diff
@@ -834,9 +865,17 @@ make_waterfall_plot <- function(dt_plot,
...
   n_obs <- length(dt_plot[, unique(id)])
-  dt_plot[, pred_label := paste0("italic(f(x)) == ", format(pred, digits = digits + 1))]
+  if (sage) {
+    dt_plot[, pred_label := paste0("italic(L(f)) == ", format(pred, digits = digits + 1))]
+  } else {
+    dt_plot[, pred_label := paste0("italic(f(x)) == ", format(pred, digits = digits + 1))]
+  }
   dt_plot[, pred_x := N_features + 0.8]
-  dt_plot[, phi0_label := paste0("~phi[0]==", format(expected, digits = digits + 1))]
+  if (sage) {
```
LHBO: Lots of if (Sage) ...  else

These could have been combined into one if else clause.

Same with line 824-825

R/setup.R
```diff
@@ -425,6 +436,9 @@ check_data <- function(internal) {
   x_train <- internal$data$x_train
   x_explain <- internal$data$x_explain
 
+  sage <- internal$parameters$scope == "global"
```
LHBO: Just want to rise the question that we are using sage internally, instead of "global". this would be more consistent. But I have no strong prefernce.

R/setup.R
```diff
@@ -728,6 +753,9 @@ check_and_set_parameters <- function(internal, type) {
 
   internal <- set_extra_comp_params(internal)
 
+  # Set the SAGE-specific parameters (loss function and baseline loss) when computing SAGE values
+  if (internal$parameters$scope == "global") internal <- set_global_parameters(internal)
```
LHBO: To match the rest of the function, sage <- internal$parameters$scope == "global" should be set in the begining of the function and then use if (sage) here

tests/testthat/_snaps/asymmetric-causal-output/output_cat_asym_causal_mixed_cat_ad.rds
LHBO, line 1: Just making a note that this file is now 120 bytes shorter. Is it not strange that it is smaller? I imagined that this branch would rather add stuff.

tests/testthat/_snaps/sage-output.md
```diff
@@ -0,0 +1,263 @@
...
+      1:          1 -1119   55.44 240.1 426.5 -23.91 -0.7152
+
+# output_sage_gaussian_xgboost
+
+    Code
+      print({
+        out <- code
+      }, digits = digits)
+    Condition
+      Warning in `throw_err_or_depr_msg()`:
```
LHBO: This might be something that we want to fix now instead of pushing it forward?

tests/testthat/_snaps/sage-output.md
```diff
@@ -0,0 +1,263 @@
...
+
+# output_sage_xgboost_binary
+
+    Code
+      print({
+        out <- code
+      }, digits = digits)
+    Condition
+      Warning in `throw_err_or_depr_msg()`:
+      Passed unrecognized parameters: verbose. This warning will become an error in a future version.
```
LHBO: Same here

vignettes/general_usage.Rmd
```diff
@@ -1647,6 +1647,92 @@ get_results(ex, what = "dt_vS")
...
+The Shapley value explanations described above are *local*: they explain a single prediction at a time.
+Sometimes we are instead interested in a *global* measure of how much each feature contributes to the overall
+predictive performance of the model. SAGE (Shapley Additive Global importancE,
+[Covert et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/c7bf0b7c1a86d5eb3be2c722cf2cf746-Paper.pdf))
+provides exactly this, by applying the Shapley value framework to a model *loss* rather than to the prediction itself.
+
+Concretely, the value of a coalition \( S \) becomes \( v(S) = -\mathbb{E}[L(y, f(x_S))] \), the (negative) expected
+loss of the model when only the features in \( S \) are known. The conditional expectations \( f(x_S) = E[f(x)\mid x_S] \)
+are estimated with the same `approach` machinery as for ordinary Shapley values, so all approaches, groups and
+iterative estimation are available for SAGE too. The resulting SAGE values are a single set of feature importances
```
LHBO, line 1661: Here you write that it works. Good.

NEWS.md
```diff
@@ -6,6 +6,7 @@
 
 ### New features
 * Added the `"arf"` approach for conditional sampling with adversarial random forests via the `arf` package ([#497](https://github.com/NorskRegnesentral/shapr/pull/497))
+* Added support for SAGE (Shapley Additive Global importancE) values via the new `scope` and `y_explain` arguments to `explain()`. Set `scope = "global"` to explain the model's global loss (using `extra_computation_args$global_loss_func`, defaulting to log-loss for binary 0/1 responses and MSE otherwise) instead of individual predictions, and plot the result with `plot()` using `plot_type = "bar"` or `"waterfall"`. (branch: sage)
```
LHBO: Same comment here regarding branch: sage

README.md
```diff
@@ -88,7 +88,10 @@ various approaches. With features like parallelized computations,
 convergence detection, progress updates, and extensive plotting options,
 shapr is a highly efficient and user-friendly tool, delivering precise
 estimates of conditional Shapley values, which are critical for
-understanding how features truly contribute to predictions.
+understanding how features truly contribute to predictions. In addition
+to local prediction explanations, `shapr` also supports global feature
+importance through SAGE (Shapley Additive Global importancE) values,
+computed by setting `sage = TRUE` in `explain()`.
```
LHBO: Here it is not updated to scope = "global".

LHBO:
```suggestion
computed by setting `scope = "global"` in `explain()`.
```

README.md
```diff
@@ -267,6 +270,72 @@ plot(explanation)
 
 <img src="man/figures/README-basic_example-1.png" alt="" width="100%" />
 
+In addition to explaining individual predictions, `shapr` can compute
+global feature importance through SAGE (Shapley Additive Global
+importancE) values by setting `sage = TRUE` and providing the observed
+responses via `y_explain`. The loss defaults to log-loss for binary 0/1
+responses and the mean squared error otherwise, and a custom loss can be
+supplied through `sage_args`.
```
LHBO: This is not correct.

README.md
```diff
@@ -267,6 +270,72 @@ plot(explanation)
 
 <img src="man/figures/README-basic_example-1.png" alt="" width="100%" />
 
+In addition to explaining individual predictions, `shapr` can compute
+global feature importance through SAGE (Shapley Additive Global
+importancE) values by setting `sage = TRUE` and providing the observed
```
LHBO: Same here. I guess you did not update this file.

README.md
```diff
@@ -267,6 +270,72 @@ plot(explanation)
...
+
+``` r
+# Compute SAGE values explaining the model's global loss (MSE by default)
+sage_explanation <- explain(
+  model = model,
+  x_explain = x_explain,
+  x_train = x_train,
+  approach = "empirical",
+  phi0 = p0,
+  sage = TRUE,
```
LHBO: Here

README.Rmd
```diff
@@ -84,6 +84,8 @@ The package also includes evaluation metrics to compare various approaches.
 With features like parallelized computations, convergence detection, progress updates, and extensive plotting options,
 shapr is a highly efficient and user-friendly tool, delivering precise estimates of conditional Shapley values,
 which are critical for understanding how features truly contribute to predictions.
+In addition to local prediction explanations, `shapr` also supports global feature importance through SAGE
+(Shapley Additive Global importancE) values, computed by setting `sage = TRUE` in `explain()`.
```
LHBO: This will be fixed when you remake the rmd files

README.Rmd
```diff
@@ -182,6 +184,31 @@ summary(explanation)
 plot(explanation)
 ```
 
+In addition to explaining individual predictions, `shapr` can compute global feature importance through SAGE
+(Shapley Additive Global importancE) values by setting `sage = TRUE` and providing the observed responses via
```
LHBO: Same here

LHBO: I will not comment anything more here about sage

LHBO (changes requested):
Good introduction of global shapley values. Nice test and checks. Sage is still included some places in stead of scope. FIx that and everything else seems nice. Well done 👍

R/setup.R
martinju: Agree, good catch. I will use is_global for consistency with is_groupwise (and global as a variable name is semi-reserved)

tests/testthat/_snaps/asymmetric-causal-output/output_cat_asym_causal_mixed_cat_ad.rds
martinju: You are correct to question this. I checked it, and there is nothing wrong -- one parameter added. AI says the reduced size is due to there being some changes in the last digits of internal numbers (10^(-15) differences) caused by two different environments where the snapshots where created, and (by chance?) these were compressed to a smaller size than the original ones.

So I think this is nothing to worry about.

tests/testthat/_snaps/sage-output.md
martinju: Oh, good catch (I shoudl really have spotted this myself!). It was caused by our use of the old xgboost interface for sage (for some unknown reason). I have updated this now.

vignettes/general_usage.Rmd
martinju: I also expanded upon this now to clarify it also works for causal and asymmetric.

NEWS.md
martinju: fix later

python/src/pyshapr/explanation.py
martinju: @LHBO 
Ok, so apparently this was correct in the first place. PEP8 (https://peps.python.org/pep-0008/#imports) says:

"Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports."

numpy is third party, while typing is from standard library, so typing shoudl come first.

Also, I could not find verification of your statement that PEP8 says "import statements always come before from ... import statements". 

Anyway, will keep it like it was so my ruff linter stays happy :-)

PR #506, opened 2026-07-18
Fix vaeac one_hot_max_sizes for all-categorical data (sapply -> lapply)

martinju:
## Summary

The `vaeac` approach fails with a torch `index ... is out of bounds` error on **all-categorical** data whose features all have the **same number of levels** (e.g. 8 factors each with 3 levels).

The cause is in `vaeac_preprocess_data()` ([R/approach_vaeac_torch_modules.R](R/approach_vaeac_torch_modules.R)):

```r
feature_list$factor_levels <- sapply(data, levels)
```

When every feature is a factor with the same number of levels, each `levels()` call returns an equal-length character vector, so `sapply()` **simplifies the result to a matrix** instead of returning a per-feature list. The subsequent `sapply(factor_levels, length)` then iterates over every matrix cell, producing a wrong `one_hot_max_sizes` of length `n_features * n_levels` (all `1`s — i.e. every feature is wrongly treated as continuous) instead of length `n_features`. `vaeac_compute_normalization()` then loops past the number of columns and overflows the data tensor:

```
index 8 is out of bounds for dimension 1 with size 8
```

Mixed data (numeric + factor, or factors with differing level counts) is unaffected, because `sapply` cannot simplify a heterogeneous result to a matrix — which is why only all-categorical data hit this.

## Fix

Use `lapply()` so the result is always a per-feature list:

```r
feature_list$factor_levels <- lapply(data, levels)
```

Confirmed on the failing data: `one_hot_max_sizes` becomes `c(3,3,3,3,3,3,3,3)` (length 8) instead of 24 ones, and mixed data is unchanged.

## Tests

- Added a fast regression test ([tests/testthat/test-vaeac-setup.R](tests/testthat/test-vaeac-setup.R)) that calls `vaeac_preprocess_data()` on all-categorical, uniform-level data and asserts `one_hot_max_sizes` has one entry per feature (equal to the number of levels), not `n_features * n_levels`.
- Full test suite: **547 passed, 5 skipped, 0 failures, 0 snapshot changes**.
- `styler` / `lintr` clean.


R/approach_vaeac_torch_modules.R
```diff
@@ -867,7 +867,10 @@ vaeac_preprocess_data <- function(data, log_exp_cont_feat = FALSE,
   feature_list <- list()
   feature_list$labels <- colnames(data)
   feature_list$classes <- sapply(data, class)
-  feature_list$factor_levels <- sapply(data, levels)
+  # Use lapply (not sapply) so the result is always a list with one entry per feature. With sapply,
+  # an all-categorical data set whose features share the same number of levels simplifies to a matrix,
+  # which corrupts the downstream one_hot_max_sizes (see vaeac_preprocess_data below).
```
martinju:
```suggestion
```

PR #517, opened 2026-09-08
Release shapr 1.1.0

martinju:
## Summary

- Prepare the `shapr` 1.1.0 release metadata and consolidate `NEWS.md`.
- Update `pyshapr` compatibility metadata, tests, and installation guidance for
  `shapr` 1.1.0.
- Regenerate the README, long-running vignettes, figures, and roxygen2 output.
- Resolve package-owned CRAN check notes and exclude `python-shim` from the R
  source package.
- Record final platform and reverse-dependency results in `cran-comments.md`.

## Validation

- Final clean-clone `R CMD check --as-cran`: 0 errors, 0 warnings, and 0 notes.
- All 5 reverse dependencies: 0 new problems and 0 packages failed to check.


python/src/pyshapr/_rutils.py
```diff
@@ -57,7 +57,7 @@ def _warn_if_shapr_version_lacks_full_support(shapr_package: Any) -> None:
         f"{_format_features(UNAVAILABLE_FEATURES)} require a shapr version newer than "
         f"{RECENT_FEATURES_REQUIRE_NEWER_THAN} and are unavailable; other "
         "functionality may still work. "
-        "Update shapr from R with pak::pak('NorskRegnesentral/shapr').",
+        "Update shapr from R with install.packages('shapr').",
```
LHBO:
```suggestion
        "Update shapr from R with `install.packages('shapr')`.",
```

python/src/pyshapr/_rutils.py
```diff
@@ -71,7 +71,7 @@ def _check_shapr_feature_support(shapr_package: Any, scope: str) -> None:
     raise ShaprVersionError(
         f"shapr {installed_version} does not support SAGE (`scope='global'`). Without this "
         "check, it can return local SHAP values instead of SAGE values. Update shapr from R with "
-        "pak::pak('NorskRegnesentral/shapr')."
+        "install.packages('shapr')."
```
LHBO:
```suggestion
        "`install.packages('shapr')`."
```

python/src/pyshapr/_rutils.py
LHBO: I think we previously did this to make it clearer that this is the code the user should run.

python/tests/setup/test_version_compatibility.py
```diff
@@ -24,19 +24,18 @@ def test_shapr_version_without_full_support_warns(monkeypatch, installed_version
     shapr_package = SimpleNamespace(__version__=installed_version)
 
     expected_message = (
-        "pyshapr 0.5.1 was developed with shapr 1.0.8.9005, but shapr "
+        "pyshapr 0.5.1 was developed with shapr 1.1.0, but shapr "
         f"{installed_version} is installed. ARF and SAGE require a shapr version newer than "
         "1.0.8 and are unavailable; other functionality may still work. "
-        "Update shapr from R with "
-        "pak::pak('NorskRegnesentral/shapr')."
+        "Update shapr from R with install.packages('shapr')."
```
LHBO:
```suggestion
        "Update shapr from R with `install.packages('shapr')`."
```

NEWS.md
```diff
@@ -2,33 +2,31 @@
...
 * Fixed `cli::cli_progress_step()` reporting artificially short timings for the v(S) computation step by passing `.envir = parent.frame()` to bind the progress bar's lifetime to the correct calling environment. ([#496](https://github.com/NorskRegnesentral/shapr/pull/496))
 
-### Tests
+
+### Development and testing
+* Added repository-wide agent instructions, VS Code tasks, and snapshot-safe local testing helpers ([#493](https://github.com/NorskRegnesentral/shapr/pull/493)).
+* Added pre-PR workflow scripts (`dev/prepare-pr`, `dev/check-pr`, `dev/publish-pr`) with `dev/pr-workflow.md` for automated and agent-assisted PR readiness checks; consolidated development scripts under `dev/`. ([#494](https://github.com/NorskRegnesentral/shapr/pull/494))
+* Added a full computational benchmark suite for runtime and peak RAM across the supported approaches and representative workloads, together with a pkgdown article presenting the retained results. The benchmark framework includes coalition-budget validation for iterative pairs, corrected process-tree RAM polling, and optional realistic parallel-workload studies. ([#510](https://github.com/NorskRegnesentral/shapr/pull/510))
```
LHBO:
```suggestion
* Added a full computational benchmark suite for runtime and peak RAM usage across the supported approaches and representative workloads, together with a pkgdown article presenting the retained results. The benchmark framework includes coalition-budget validation for iterative pairs, corrected process-tree RAM polling, and optional realistic parallel-workload studies. ([#510](https://github.com/NorskRegnesentral/shapr/pull/510))
```

cran-comments.md
```diff
@@ -1,35 +1,61 @@
...
+Both checks excluded snapshot tests without explicitly passing `--as-cran`.
+
+* Local Ubuntu 24.04.4, R 4.6.1 (release): depends-only check without suggested packages
+* R-hub Clang ASAN on Ubuntu 22.04.5, R devel: CRAN-like check with suggested packages
 
 ## Current R CMD check results
 
-There were no ERRORs, WARNINGs, or NOTES.
+There were no ERRORs or WARNINGs, and all snapshot tests passed locally and on
+GHA. The local `R CMD check --as-cran` produced one NOTE caused by the
```
LHBO: Should this be fixed? I.e., that we get no note. Note sure what this flag means.

vignettes/asymmetric_causal.Rmd
```diff
@@ -458,35 +460,44 @@ explanation_sym_con[["gaussian"]] <- explain(
 #>
 #> • Number of observations to explain: 144
 #>
-#> • Computations (temporary) saved at: '/tmp/RtmplCNqcR/shapr_obj_39ed0d15682a5b.rds'
+#> • Computations (temporary) saved at:
```
LHBO: Just a comment. I assume the reason for a new line is now that we have a max width.

vignettes/asymmetric_causal.Rmd
```diff
@@ -595,9 +621,9 @@ All approaches are comparable, but `xgboost` is clearly the fastest approach.
 ``` r
 print_MSEv_scores_and_time(explanation_sym_con)
 #>             MSEv MSEv_sd Time (secs)
-#> gaussian 1101051   71271       12.17
-#> ctree    1060934   63108       86.50
-#> xgboost  1090931   58988       18.60
+#> gaussian 1093465   71390       11.73
```
LHBO: The `gaussian` approach have now changed a bit due to each coalition is now being split into batches to ensure that RCPP/Rarmadillo do not use too much RAM. This cause the seed/randomness to not match.

vignettes/asymmetric_causal.Rmd
```diff
@@ -1786,41 +1873,60 @@ explanation_n_coal[["sym_cau_gaussian_64"]] <- explain(
...
-#> ── Iteration 4 ──────────────────────────────────────────────────────────────────────
 #>
-#> ℹ Using 50 of 128 coalitions, 4 new.
 #>
+#> ── Iteration 5 ─────────────────────────────────────────────────────────────────
 #>
+#> ℹ Using 58 of 128 coalitions, 6 new.
 #>
-#> ── Iteration 5 ──────────────────────────────────────────────────────────────────────
+#> ℹ Coalitions split into 6 batches (mean 9.7 per batch).
```
LHBO: Note too my self. Check how this mean value is computed. 
We are now in iteration 5 and using 58 coalitions, and 6 are new. -> How did we get 9.7 per batch.

vignettes/general_usage.Rmd
```diff
@@ -441,6 +441,11 @@ library(shapr)
 ``` r
 library(xgboost)
 library(data.table)
+#>
+#> Attaching package: 'data.table'
+#> The following object is masked from 'package:base':
```
LHBO, line 446: This is because we now use R 4.5+.

Can `suppressPackageStartupMessages` if we want to suppress the message or use ```{r, message=FALSE}``` to make knitr not show it. I.e., in the header of the code chunk.

vignettes/general_usage.Rmd
```diff
@@ -1211,11 +1253,13 @@ explanation_gaussian_1e2 <- explain(
...
 #> • Number of feature-wise Shapley values: 4
 #> • Number of observations to explain: 25
-#> • Computations (temporary) saved at: '/tmp/RtmpIOXT2V/shapr_obj_329d9f2f0f8c3f.rds'
+#> • Computations (temporary) saved at:
+#> '/tmp/RtmpuzSVRD/shapr_obj_8cd3d788f0dd9.rds'
 #>
 #> ── Main computation started ──
 #>
 #> ℹ Using 16 of 16 coalitions.
+#> ℹ Coalitions split into 10 batches (mean 1.6 per batch).
```
LHBO, line 1132: Here the mean batch makes sense. 16 coalitions are split into 10 batches -> 1.6

vignettes/regression.Rmd
```diff
@@ -249,39 +249,42 @@ explanation_list$MC_empirical <- explain(
...
+#> ℹ `max_n_coalitions` is `NULL` or larger than `2^n_features = 16`, and is
+#>   therefore set to `2^n_features = 16`.
+#>
+#>
 #> ── Explanation overview ──
-#> 
-#> 
-#> 
+#>
+#>
```
LHBO: Also look at how the vignettes look like. Do they actually include this many lines of white space? Usually, it was just one new line.

LHBO (commented):
Solid update and thing seems correct.

Just added some minor suggestions.

Note that figures have changed a bit. Unsure why as we set a seed, but this is likely caused by the new batching of coalitions, which means that the sampling process is not 1-to-1 with the previous version even with the same seed.
Must check the avg number of coalitions per batch for the iterative version. 
I will approve when I have understood this part.

Also there has been some pep 8 corrections with trailing commas and minimum two white spaces before inline comment.

vignettes/asymmetric_causal.Rmd
LHBO: The relevant code is in `cli.R`:
```
    n_batches <- length(internal$iter_list[[iter]]$S_batch)
    n_coalitions <- internal$iter_list[[iter]]$n_coalitions
    mean_batch_size <- round(n_coalitions / n_batches, 1)
    cli::cli_alert_info("Coalitions split into {.val {n_batches}} batches (mean {.val {mean_batch_size}} per batch). ")
```

vignettes/regression.Rmd
LHBO: Yes, they do. Not a big problem. Probably a reason why it happens in the vignettes but not when run in the console/terminal.

vignettes/asymmetric_causal.Rmd
LHBO: Forget it. I understand that it, it is across all precious iterations. But I still feal like it is a bit incorrect.
Running the example in the main vignette we get:
 ```
 # Explanation with iterative computation
ex <- explain(
  model = model,
  x_explain = x_explain,
  x_train = x_train,
  approach = "gaussian",
  phi0 = p0,
  seed = 1,
  iterative = TRUE,
  iterative_args = list(convergence_tol = 0.1)
)
── Iterative computation started ──

── Iteration 1 ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
ℹ Using 6 of 16 coalitions, 6 new. 
ℹ Coalitions split into 4 batches (mean 1.5 per batch). 

── Iteration 2 ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
ℹ Using 8 of 16 coalitions, 2 new. 
ℹ Coalitions split into 2 batches (mean 4 per batch). 

── Iteration 3 ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
ℹ Using 10 of 16 coalitions, 2 new. 
ℹ Coalitions split into 2 batches (mean 5 per batch).
```

Let us discuss iteration 3, we have done a total of 10 coalitions over 3 iterations.  And IN THE THIRD iteration, we use only 2 batches and then say that the mean is 10 / 2 = 5. 

But this does not take into consideration that we used 4 batches in Iteration 1 and 2 in iteration 2. So we have used a total of 8 batches to compute the 10 coalitions. So a more correct value is 10 / 8.

Or, the mean per batch should be focused only on the latest iteration. So in the third iteration, it should be 2 new coalitions and 2 batches -> mean 1 per batch. 

What do you think Martin?

Comment (LHBO) [reply vignettes/asymmetric_causal.Rmd]: So for the latter we would replace `internal$iter_list[[iter]]$n_coalitions` with `internal$iter_list[[iter]]$new_n_coalitions`.

File: vignettes/regression.Rmd (hunk 11 above)
Comment (LHBO) [reply vignettes/regression.Rmd]: A possible solution is to extend `strip_output_trailing_whitespace()` in `dev/rebuild-long-running-vignettes.R` with

OLD:
``` 
strip_output_trailing_whitespace <- function(file) {
  lines <- readLines(file, warn = FALSE)
  fixed_lines <- lines
  changed <- startsWith(lines, "#>") & grepl("[[:space:]]+$", lines)

  if (any(changed)) {
    fixed_lines[changed] <- sub("[[:space:]]+$", "", fixed_lines[changed])
    writeLines(fixed_lines, file)
  }

  return(sum(changed))
}
```

NEW (with new function name):
``` 
clean_output_lines <- function(file) {
  lines <- readLines(file, warn = FALSE)

  # Remove trailing whitespace from output lines ("#>  " -> "#>")
  lines <- sub("^(#>.*?)[[:space:]]+$", "\\1", lines)

  # Remove empty output lines ("#>" are not kept")
  keep <- !grepl("^#>[[:space:]]*$", lines)
  
  # Write the files we are to keep
  writeLines(lines[keep], file)
  
  # Return the number of changed sentences
  return(sum(!keep))
}
```
I.e., replace `n_trailing_whitespace_fixed <- strip_output_trailing_whitespace(output)` with `n_trailing_whitespace_fixed <- clean_output_lines(output)` later in the same file.

Note that then the output message after this call also has to been update to reflect that the count is both the number of trailing whitespaces and "empty" lines.  Or can have two separate functions to log both of these values.

vignettes/asymmetric_causal.Rmd
martinju: Good catch! It was intended to be the current one only. My tests when implementing this were apparently too weak. Will update to use new_n_coalitions as you propose.

cran-comments.md
martinju: TODO: Remove (after a check that)

